PR #8, opened 2016-10-19
Fix build on gcc 6.2

est31:
It gave me a build error when I tried to compile it on gcc 6.2.

Also, if you are interested, this is a list of compilation warnings I've encountered with the compiler version: http://pastebin.com/W0rXwnWF

Just mentioned those as I could remember mrustc didn't do any warnings back when I was on the older gcc version.


thepowersgang:
Looks good. Not sure why I didn't need `<numeric>` locally...

PR #9, opened 2016-12-24
Makefile - sed no longer uses extended regex

ubsan:
This means that it now works on both Linuxes and BSDs.


ubsan:
I'm going to do a whole PR making it work on MacOS.

PR #14, opened 2016-12-25
Fix the build on macOS

### LINKED ISSUE #13: Doesn't compile on macOS
`sed -r` doesn't work, and for some reason `FMT` doesn't work either.

ubsan:
basically three changes:

1) cast to uint64_t

2) change `sed` to no longer use `-r`

3) `FMT` macro becomes a function `format` (this is the big one)

fixes #13

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,10 @@ output/lib%.hir: $(RUSTCSRC)src/lib%/src/lib.rs $(RUSTCSRC) $(BIN)
 
 fcn_extcrate = $(patsubst %,output/lib%.hir,$(1))
 
-fn_getdeps = $(shell cat $1 | sed -nr 's/.*extern crate ([a-zA-Z_0-9]+)( as .*)?;.*/\1/p' | tr '\n' ' ')
+fn_getdeps = \
+  $(shell cat $1 \
+  | sed -n 's/.*extern crate \([a-zA-Z_0-9][a-zA-Z_0-9]*\)\( as .*\)\{0,1\};/\1/p' \
+  | tr '\n' ' ')
 
 output/libarena.hir: output/libcore.hir
 output/liballoc.hir: output/libcore.hir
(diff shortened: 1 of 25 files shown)


est31:
Maybe also edit the .travis.yml file to do builds on macOS too? Then it won't regress in the future (mutahbah keeps a close look on travis afaik).

ubsan:
This is unmergeable right now. I need to figure out what's going wrong with tests...

est31:
@ubsan

```
$ export STR="extern crate serialize as rustc_serialize; // used by deriving"
$ echo "$STR"  | sed -n 's/.*extern crate \([a-zA-Z_0-9][a-zA-Z_0-9]*\)\( as .*\)\{0,1\};/\1/p'
serialize // used by deriving
$ echo "$STR"  | sed -nr 's/.*extern crate ([a-zA-Z_0-9]+)( as .*)?;.*/\1/p'
serialize
```

If you add a `.*` to the end of your regex it seems to be fixed.

ubsan:
I need to continue to attempt to fix this; `make test` currently silently fails on macOS. However, it works for Linux and Windows, apparently, so I think it'd be nice to merge this patch in the interim. At least `make` now works!

thepowersgang:
I'm inclined not to merge a change that alters so much of the codebase along with relatively minor fixes. Sorry if the `FMT` macro bothers you.

ubsan:
@thepowersgang Modern C++ should not use macros at all. Modules, in their current incarnation, don't even support them. The C++ core guidelines also recommend against them (here)[https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-macros].

ubsan:
Alright, from talking on IRC, I'm going to split `sed` and `FMT` into two commits, as well as start getting rid of the other macros (specifically, starting with the `TU_*` macros).

PR #15, opened 2016-12-27
Fix sed to work across unices

ubsan:
This is simply the `sed` fix, starting to get it to work on macOS.


thepowersgang:
Thank you very much :)

PR #16, opened 2017-01-07
Fix build

est31:
Build errored for me with:

```
src/trans/codegen_c.cpp:1129:37: error: ‘isnan’ is not a member of ‘std’
                                 if( ::std::isnan(c) ) {
src/trans/codegen_c.cpp:1132:42: error: ‘isinf’ is not a member of ‘std’
                                 else if( ::std::isinf(c) ) {
```

now it does not anymore.


thepowersgang:
Thanks.

PR #48, opened 2018-01-07
TestRunner - Add option --fail-fast, don’t fail fast by default

jirutka:
Note: I’m not C++ developer, so I just _hope_ that I coded it right.


thepowersgang:
Nice idea :) But brace style isn't quite correct.

jirutka:
Is it better now?

kaniini:
it looks fine to me now

thepowersgang:
I didn't actually see any new commits after your comment... but meh, it'll probably get cleaned up in the next refactor.

jirutka:
I've amended it to the original commit and force pushed.

PR #67, opened 2018-03-21
Update MIR.md

aristarh2704:


thepowersgang:
I appreciate the attempt, but that's not correct. They're "PARAM" not "RVALUE"

PR #68, opened 2018-03-26
netbsd/amd64 support

coypoop:
With some blunt changes to Makefiles which I did not include, it build std and it seems to work. Let me know if I did anything wrong.

(Random notes: along the way I ran into the GCC 5.5 issue (66) and a GCC dual ABI one, and avoided those by using a non-dual-ABI GCC 6.4)


thepowersgang:
Looks good, thanks.

PR #71, opened 2018-04-01
Add targets aarch64-linux-gnu and aarch64-unknown-openbsd 

jirutka:
Note: I haven’t tested if it actually work on aarch64 yet.

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -24,6 +24,11 @@ const TargetArch ARCH_X86 = {
     32, false,
     { /*atomic(u8)=*/true, false, true, false,  true }
 };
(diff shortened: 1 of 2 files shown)


src/trans/target.cpp
```diff
@@ -24,6 +24,11 @@ const TargetArch ARCH_X86 = {
     32, false,
     { /*atomic(u8)=*/true, false, true, false,  true }
 };
+const TargetArch ARCH_ARM64 = {
+    "aarch64",
+    64, false,
+    { /*atomic(u8)=*/true, true, true, true,  true }
```
jirutka: I’m not sure about atomic u16.

Issue #87, opened 2018-10-20
Building on Windows - CMake?


ataulien:
I tried to compile mrustc a couple of times now for windows (my work machine), but unfortunately I did not succeed.

## Stuck on VS2015's compiler

The build only works with the Compiler from VS2015 (v140), which is what is set inside the projects. When switching the setting to VS2017 (v141) or Clang, the build is missing zlib, which is supposedly pulled in via NuGet. It seems that there is no zLib-package on NuGet for v141 so the build just fails when including a zlib-header.

I am not too experienced with NuGet on C++ and I could not find a workaround for that (manually downloading zlib would work I suppose).

## Wrong directories for 32-bit target

When trying to compile as 32-bit, VS complains about not finding `common.lib`, even though it is being built. The problem here is that the library output location for a 32-bit build of `common.lib`  does not match the location of the 64-bit build.

## Unresolved externals to `HIR::ExprPtr::get_mir_or_error_mut` and others

There is a call to `HIR::ExprPtr::get_mir_or_error_mut` in `cleanup.cpp`, inside `MIR_CleanupCrate`. The linker complains about it not finding `get_mir_or_error_mut`, which I could not find the reason for. 
I can grab the exact build error when I'm at work again next week, but I remember VS looking for a function with `__cdecl`-calling convention, when it should rather be a `__thiscall`... 

## No CI for Windows

The excellent AppVeyor provides CI for Windows builds free of charge for Open Source-projects. Is there any reason this has not been set up yet other than "nobody bothered yet"? If the latter is true, I would go ahead and set it up if I can get the windows build going on my own machine.

## Switch to CMake?

Right now, the project ships with makefiles and a Visual Studio solution. If you think it would be worth a try, I would set up CMake for mrustc to see if it would make building easier for all platforms. 

If you have never worked with CMake, here are some features which I think could help the project from the top of my head:

 - One of the most used build-systems in the C/C++ world these days
 - Not depending on any IDE or compiler
 - Cross-platform downloading files with checksum-tests
 - Streamlines build process on each platform
 - Can generate "install"-targets which makes packaging easy

ubsan:
I think Meson would also be a good choice to think about. I'd be totally willing to set up a PR for compiling it this way.

One thing I'd like to remind people is that all of these options are installable with `pip`.

thepowersgang:
[...]

PR #88, opened 2018-10-21
[WIP] Build via CMake

ataulien:
See issue #87. 

## This is WIP - Don't merge yet!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
  - cmake ..
  - make mrustc -j4
  - make minicargo -j4
- - make rust_standard_libs -j4
+ # - make rust_standard_libs -j4
 
 # Test
 # - make -f minicargo.mk output/libtest.hir      # libstd


dongcarl:
@ataulien What is the progress with this? It seems like the checks are passing? Anything more to do?

dongcarl:
@ataulien I think as long as `mrustc` and `minicargo` builds fine with cmake, there could be a separate PR for the rest?

Just wondering what's missing and how I could help.

ataulien:
@dongcarl It seems to just hand at some point while ninja runs. I'll have a look at it tomorrow or so.

ataulien:
"Tomorrow" is today and I got the build working on travis. Turns out using `make` instead of `ninja` and getting rid of `set -e` makes it work.

It builds `mrustc` and `minicargo` just fine, but still fails at compiling the rust std-libraries.

thepowersgang:
If you make `rust_libpanic_unwind` and `rust_libtest` depend on `rust_libstd` then the standard library should compile.

Also, `libproc_macro` is a custom crate, it source is in `lib/libproc_macro`

rdrpenguin04:
The referenced issue has been closed; is this needed anymore?

thepowersgang:
Probably not needed? I feel that cmake is a little too much complexity to manage over just having a vsproject and makefile.

orowith2os:
I think I'll note that I would *love* to see meson - makes it easier to integrate in some build system stuff I tend to tinker with (Flatpak, freedesktop-sdk)

thepowersgang:
> I think I'll note that I would _love_ to see meson - makes it easier to integrate in some build system stuff I tend to tinker with (Flatpak, freedesktop-sdk)

If mrustc gets a unified build system, it'd most likely be cmake - because cmake is already needed for llvm, and because I use it for my day job.

PR #89, opened 2018-10-21
[WIP] Build with meson

ubsan:
Since I'm a VS2017 user, I also made sure to get it compiling with VS2017.

This is just an initial pass. I think it's a useful comparison with #88, as I think meson is far simpler and easy to read compared to CMake.


ubsan:
You'll also notice that the versions update every build, as opposed to every run of CMake, like in #88.

thepowersgang:
Could you separate the source changes out from the new build system?

thepowersgang:
I'm closing this PR in favour of using CMake (sorry).
However, I'd appreciate the VS2017 fixes in a separate PR.

ubsan:
np! I'll see if I can do the thing.

PR #90, opened 2018-10-27
Parameterize rust version

ericherman:
This pull-request is a step towards making the build scripts a little less fragile regarding version.

It seems that 1.19.0 is no longer hosted at https://static.rust-lang.org/dist/rustc-1.19.0-src.tar.gz and the TestRustcBootstrap.sh was altered to fetch 1.20.0; however the build still has 1.19 hardcoded in many places.

(The vsproject files still have 1.19.0, and thus are not yet addressed.)

### DIFF
--- a/TestRustcBootstrap.sh
+++ b/TestRustcBootstrap.sh
@@ -8,11 +8,22 @@ WORKDIR=${WORKDIR:-rustc_bootstrap}/
 MAKEFLAGS=-j8
 export MAKEFLAGS
 
+if [ -z ${RUSTC_SRC_DES} ]; then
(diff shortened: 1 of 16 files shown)


thepowersgang:
1.19 is the supported target (1.20 won't build)
The bootstrap script uses 1.20 intentionally (it uses the mrustc-built rustc 1.19 as the stage0 in place of a downloaded 1.19). So, that script should always be using one higher version than the current target.

The link for rustc-1.19.0-src.tar.gz works for me.

ericherman:
Okay. Now I understand. Thank you.

Issue #83, opened 2018-09-08
armhf - alignment and sizes


daym:
This issue is to collect the results I have when finding out the alignment of struct members on armhf.

The following program has been run on armhf:
```
#define _GNU_SOURCE
#include <stdio.h>
#include <stdint.h>
#include <stddef.h>

struct foo {
        unsigned char a;
        uint16_t b;
};

int main() {
        struct foo f;
        printf("b_offset = %zu\n", offsetof(struct foo, b));
        return 0;
}
```

This results in: b_offset = 2.

Then I modified it accordingly to get:

```
Type         Alignment/Bytes
int8_t       1
int16_t     2
int32_t     4
int64_t     8
uint8_t    1
uint16_t   2
uint32_t   4
uint64_t   8
__int128  not supported on target
float         4
double    8
long double 8   (size = 8)
```

glaubitz:
m68k (gcc 8.2.0):

```
Type         Alignment/Bytes  Size/Bytes
int8_t       1                1
int16_t      2                2
int32_t      2                4
int64_t      2                8
__int128     not supported
uint8_t      1                1
uint16_t     2                2
uint32_t     2                4
uint64_t     2                8
float        2                4
double       2                8
long double  2                12
size_t       2                4
off_t        2               4
void*        2               4
```

eddyp:
Does it make sense to have the alignments and sizes for other compilers than gcc/llvm?

daym:
I don't know, you tell me. If the libraries these compilers create interoperate with libraries other compilers create then those alignments and sizes have to be equal between compilers anyway.

glaubitz:
Alignment sizes are dictated by the ABI which is usually independent of the compiler used.

See for example: https://wiki.osdev.org/System_V_ABI

glaubitz:
But wouldn't that make any code compiled with a different compiler ABI-incompatible with standard Linux environments, for example?

eddyp:
It would. But I mostly thinking of bare metal code and embedded systems where the problem is not that obvious and using the same compiler is the go to solution.

glaubitz:
Okay, but bare-metal code doesn't use the SysV ABI anyway, so we're talking about different things then.

PR #92, opened 2018-11-14
Trans - Add initial support for m68k

glaubitz:
This adds initial support for m68k, currently on Linux only. With these changes, I am able to build mrustc on qemu-m68k.

There are some alignment issues that need to be sorted out due to the fact that the natural alignment on m68k is 16 bits and not 32 bits, despite being a 32-bit architecture. I will address those issues in future patches.

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -34,6 +34,11 @@ const TargetArch ARCH_ARM32 = {
     32, false,
     { /*atomic(u8)=*/true, false, true, false,  true }
 };
+const TargetArch ARCH_M68K = {
+    "m68k",
+    32, true,
+    { /*atomic(u8)=*/true, false, true, false,  true }
+};
 TargetSpec  g_target;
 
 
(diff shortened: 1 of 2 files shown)


daym:
Alignment is handled in trans/target.cpp Target_GetSizeAndAlignOf.

There's already a case analysis there for "x86", so you can just add yours.

See https://github.com/thepowersgang/mrustc/issues/83 for a C program to run on your m68k to quickly find out the alignments.

glaubitz:
Thanks, this is very helpful! I will look into this!

glaubitz:
Sorry, I was very busy this week. I wanted to have a go at it this weekend and if I don't find the time, it will be over the holidays.

I have also started packaging ```mrustc``` for openSUSE and will eventually package it for Debian, too.

Boscop:
Any update on this? :)

I'm very much looking forward to using mrustc to compile Rust code to C for my TI Voyage 200 calculator which has a m68k CPU (using the TIGCC compiler).

PR #94, opened 2018-11-29
Mark destructors on base classes as virtual

arlied-google:
In many situations, objects are deleted by deleting a pointer to their base class. In the previous code, these destructors were non-virtual. This violates a C++ rule (so it's undefined behavior), and on some platforms (such as Clang with -fsized-deallocation), this violation is detected and causes mrustc to abort.

The fix is simple; mark the relevant destructors as virtual.

### DIFF
--- a/src/ast/expr.hpp
+++ b/src/ast/expr.hpp
@@ -623,6 +623,7 @@ struct ExprNode_UniOp:
 class NodeVisitor
 {
 public:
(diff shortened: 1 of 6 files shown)


thepowersgang:
Looks good, thanks :)

arlied-google:
Thanks!

PR #96, opened 2018-11-30
This adds a Bazel package (the BUILD file), which allows you to build mrustc using Bazel.

arlied-google:
Bazel is a powerful, high-performance build system
with a strong emphasis on correctness and reproducible builds.

This also adds zlib as a submodule. This simplifies building
dependencies. To build from scratch:

git submodule init
git submodule update --checkout
blaze build :mrustc

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "zlib"]
+	path = zlib
+	url = https://github.com/madler/zlib
--- a/BUILD
+++ b/BUILD
@@ -1,0 +1,423 @@
+# This is a Bazel package. See: https://bazel.build/ . Bazel is a
+# high-performance (parallel, cached) build system, with a strong focus on
+# correct / reproducible builds.
+#
+# This package also builds zlib, which is a Git submodule of this Git
+# repository. If this target fails to build, then make sure you have initialized
+# and updated submodules, e.g.:
+#
+#   git submodule init
+#   git submodule update
+#   bazel build :mrustc
+#
+#
+# To build with Clang:
+#
+#    bazel build --client_env=CC=clang ...
+#
+# To build with GCC:
+#
+#    bazel build --client_env=CC=gcc ...
+#
+# If you do not specify a compiler to use, Bazel will search your system for a
+# reasonable default toolchain. If you're feeling ambitious, you can also use
+# Bazel's CROSSTOOL support.
+
+GCC_COPTS = [
+        "-Wno-misleading-indentation",
+]
+
+CLANG_COPTS = [
+        # These options are specific to clang++. g++ will report a warning,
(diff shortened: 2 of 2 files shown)


thepowersgang:
See #87 for a discussion of adding a meta-build system. My current opinion is to use cmake as it reduces the maintenance and dependency overheads (as llvm via rustc already needs cmake).

arlied-google:
I see.  Well, I'd still encourage you to take this change, for these reasons:

1) There's no CMake generator for Bazel.
2) Bazel is amaaaaaazing, and I'd much rather work with Bazel than CMake.
3) (selfish) I'm working on importing mrustc into Google, for some work we are doing on Rust at Google, and our internal build system is basically == Bazel, so it would be great to have a Bazel BUILD description maintained directly in mrustc's tree.
4) It's just a single file, so it has very little impact for those are aren't using Bazel.
5) Did I mention how great Bazel is?  Works on Windows and Mac, too.

thepowersgang:
1. Does this bar Bazel from calling into cmake?
2. I appreciate that you like it, but is that a good enough reason to add it as a dependency for all other users of this project?
3. Could such a file be auto-generated from the cmake files? That would reduce the maintenance impact for mrustc while still integrating with your build system.
4. A single file listing all of the files in the project, which would need to be kept up to date.
5. So does cmake.

(Sorry if this seems snippy)

arlied-google:
Bazel calling into CMake would not work. It would entirely defeat the purpose of using the parallel analysis + scheduling of Bazel.

"Could such a file be auto-generated from the cmake files?"  That's what a CMake generator is, and such a generator does not currently exist.

It's your project, of course.  If you think the maintenance burden of a BUILD file is not worth it, then obviously I'll just have to keep that in my fork of the project.

Nothing is snippy, so long as it is about the work.

emosenkis:
In the directions, I think you meant 'bazel', not 'blaze'.

At a glance, it looks like you could use globs to avoid listing all the files. I'm not sure if that would be sufficient to avoid being a maintenance burden.

daym:
For what it's worth, Bazel requires a lot of Java packages. Those are rather well-known not to take bootstrappable nor reproducible builds seriously. So this entire thing would depend on Java for no reason other than make not being parallel enough (how exactly?), increasing the build dependency graph by several hundred packages.

There's nothing wrong with GNU Make, especially for projects like this.

arlied-google:
There's quite a bit wrong with GNU Make. Bazel provides a much more reliable, reproducible build environment than GNU Make does.

I've abandoned this work, though, because the repo owner didn't want it in the tree.

thepowersgang:
Closing after the above comment, see #87 for more discussion on non-make build systems.

PR #97, opened 2018-12-03
Fix a few bugs (in non-debug config), clean up warnings

arlied-google:
In several places, we called functions and relied on their
side-effects, within an assert() call. This means that when
building for non-debug configurations, the side-effect
expressions did not run, and so they never computed the
values that we needed (stuff we needed even beyond the
assert() call). This fixes them by adding an assert_or_ignore()
function, an ignore() function, and by moving some repeated code
to a function in codegen_mmir.cpp.

This also fixes several warnings. GCC and Clang do control-flow
analysis slightly differently. In some situations, GCC thinks that
a particular point in control-flow is reachable, while Clang thinks
that the same point is not reachable. This causes the analysis of
whether a function always returns a value to fail (erroneously).
This commit fixes this by removing a few "break;" statements that
were confusing GCC.

### DIFF
--- a/src/hir_expand/annotate_value_usage.cpp
+++ b/src/hir_expand/annotate_value_usage.cpp
@@ -61,7 +61,7 @@ namespace {
             root_ptr->m_usage = this->get_usage();
             auto expected_size = m_usage.size();
             root_ptr->visit( *this );
-            assert( m_usage.size() == expected_size );
+            assert_or_ignore( m_usage.size() == expected_size );
         }
         void visit_node_ptr(::HIR::ExprNodeP& node_ptr) override
         {
@@ -75,7 +75,7 @@ namespace {
 
             auto expected_size = m_usage.size();
             node_ptr->visit( *this );
-            assert( m_usage.size() == expected_size );
+            assert_or_ignore( m_usage.size() == expected_size );
         }
 
         void visit(::HIR::ExprNode_Block& node) override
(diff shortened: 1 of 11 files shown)


thepowersgang:
Thanks for the PR, but I feel that this is not quite suitable for merging.

- You appear to have replaced all asserts with the new  `assert_or_ignore` call (which is actually less useful than `assert` because it loses the assertion message.
- You've mixed functional changes (the assert changes) with non-functional changes (the warning clean-ups) - I prefer to keep functional and non-functional changes separate wherever possible


I do like the idea of `Target_GetSizeOf_Required`, but maybe it'd be better to use `ASSERT_BUG` instead of an assert there.

arlied-google:
OK.  How would you like this done?  One PR with a sequence of changes, with functional changes separated from warning cleanups?  At some point, super-fine-grained changes stop becoming worth the overhead.

thepowersgang:
Preference is for separate commits to separate functional and non-functional changes (but in the same PR is ok).

thepowersgang:
@arlied-google Any progress on updating the PR?

Also, I sent an email a week ago, but you can also find my email address (many of them) in the commit history.

thepowersgang:
Sorry, I've been on a programming spree today, and went and fixed the assert side-effects in the above commit. The rest of the PR could still be useful, but you may want to re-make it without the assert changes.

PR #112, opened 2019-02-07
Implement compiler check TODO.

reitermarkus:


reitermarkus:
The AppVeyor failure doesn't seem related.

thepowersgang:
Thanks for the PR.

Looking at the ordering of those statements, it may be prudent to move the check/call of <target>-gcc to before the check of the CC environment variable. This doesn't block merging of this change, but if you want to do so, go ahead.

reitermarkus:
Done.

PR #127, opened 2019-11-19
Fix typo previously breaking nightly build

E5ten:
cd to $(RUSTCSRC) not $(RUSTSRC)


thepowersgang:
Thanks... but I'm curious as to why you're trying to build a nightly :)

PR #134, opened 2020-01-03
Expand - add support for invoking procedural macros on Windows

arlosi:
This change adds support for invoking procedural macros on Windows using Win32 APIs.

### DIFF
--- a/src/expand/proc_macro.cpp
+++ b/src/expand/proc_macro.cpp
@@ -15,9 +15,9 @@
 #include "proc_macro.hpp"
 #include <parse/lex.hpp>
 #ifdef _WIN32
+# define NOMINMAX
 # define NOGDI  // Don't include GDI functions (defines some macros that collide with mrustc ones)
 # include <Windows.h>
-# undef min
 #else
 # include <unistd.h>    // read/write/pipe
 # include <spawn.h>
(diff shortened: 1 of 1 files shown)


src/expand/proc_macro.cpp
```diff
@@ -163,15 +163,15 @@ struct ProcMacroInv:
...
     HANDLE  child_handle;
     HANDLE  child_stdin;
     HANDLE  child_stdout;
 #else
     // POSIX
     pid_t   child_pid;  // Questionably needed
-     int    child_stdin;
-     int    child_stdout;
+    int    child_stdin;
+    int    child_stdout;
```
bjorn3:
```suggestion
    int     child_stdout;
```

thepowersgang: You're pointing out the change in indentation? I assume it was visual studio getting oppinionated.

thepowersgang:
Thanks for this, a few quick comments (none of which are blockers for merging really...)

 - Indentation changes around the unix handling (pretty minor really) - not really the scope of the change.
- I don't like using brace initialisation for primitive like values (the `DWORD bytesWritten{};`), kinda perfer using assignment of a value.

Feel free to argue for the current style.

src/expand/proc_macro.cpp
bjorn3: Yes, with this change it is aligned again.

arlosi: I reverted the whitespace changes

PR #135, opened 2020-01-06
minicargo - Allow build.rs to specify dependencies with '-'

arlosi:
Build dependency lines containing a `-` character were ignored. This caused breaks when the dependency line contained a path to a directory with a `-` in it (such as `output-1.29.0`).

This change resolves the issue by removing the check for `-`. Removing the check didn't appear to break anything, so I'm unsure of the intent of the check.

### DIFF
--- a/tools/minicargo/manifest.cpp
+++ b/tools/minicargo/manifest.cpp
@@ -890,7 +890,7 @@ void PackageManifest::load_build_script(const ::std::string& path)
             }
             // - Ignore
             else {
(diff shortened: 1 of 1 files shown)


thepowersgang:
Likely it was intended to check `key` instead of `line`, as `-` isn't valid in environment variables

arlosi:
I've updated the PR to check the `key` for a `-`.

thepowersgang:
Oops, build failure - the `string_view` clone doesn't have that method.

arlosi:
Oops, changed to use `std::find` instead. Build succeeds now.

PR #136, opened 2020-01-08
codegen_c - Fixes for MSVC intrinsics and emulated 128 bit math.

arlosi:
- __builtinpopcount returned the wrong result
- Overflowing multiplication functions failed to handle the case where the multiplicand was zero (causing divide by zero).
- Overflowing subtraction and addition functions did not correctly handle signed types.
- 128 bit signed right shift produced the wrong result for integers with the upper half set.
- 128 bit literals were emitted without wrapping in the struct, causing c compilation errors.
- Casts to 128 bit values did not extend the sign of the source.

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -322,8 +322,8 @@ namespace {
                 break;
             case Compiler::Msvc:
                 m_of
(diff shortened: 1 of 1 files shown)


thepowersgang:
Nice work :) I'll trust that it works (because I don't test msvc builds often - as you can tell)

PR #150, opened 2020-04-02
codegen_c - use aggregate initialization instead of GCC statement expressions

arlosi:
This change uses aggregate initialization when initializing constants instead of GCC statement expressions.

Initialization of statics that are zero is skipped, since they will be zero-initialized on program startup anyway. This allows zeroed statics with more than 16 levels of nesting to be compiled with out errors on MSVC.

If a static has more than 16 levels of nesting, and is not zero-initialized, MSVC will fail with:
`C1054: compiler limit: initializers nested too deeply`

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -2004,12 +2004,15 @@ namespace {
             TRACE_FUNCTION_F(p);
 
             auto type = params.monomorph(m_resolve, item.m_type);
-            emit_ctype( type, FMT_CB(ss, ss << Trans_Mangle(p);) );
-            m_of << " = ";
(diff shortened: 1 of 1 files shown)


arlosi:
Skipping initialization of zeroed statics is a bit of a hack, but this was the last major issue in getting rustc and cargo compiling on Windows.

thepowersgang:
Not sure why Travis isn't running on this, but I'll merge and check manually.

thepowersgang:
Seems to work :)

bjorn3:
> Not sure why Travis isn't running on this

This is not the only repo Travis isn't working on.

PR #153, opened 2020-04-30
Add support for link_name attribute on statics

arlosi:
Rust supports placing the link_name attribute on statics in addition to functions.

This functionality is used in rustc itself (though only on Windows) in `libpanic_unwind/seh.rs` as
```
    #[link_name = "\x01??_7type_info@@6B@"]
    static TYPE_INFO_VTABLE: *const u8;
```

To support this usage, this change also uses the linker alternatename option instead of #define when emitting C code on MSVC.

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -1993,7 +1993,14 @@ namespace {
             m_mir_res = &top_mir_res;
             TRACE_FUNCTION_F(p);
 
-            if( item.m_linkage.name != "" && m_compiler != Compiler::Gcc )
+            // LLVM supports prepending a symbol name with \1 to prevent further mangling.
+            // Since we're targeting C, not LLVM, strip off this prefix.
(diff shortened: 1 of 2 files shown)


src/trans/codegen_c.cpp
```diff
@@ -1993,29 +1993,35 @@ namespace {
...
+            }
+
+            if( linkage_name != "" && m_compiler != Compiler::Gcc )
             {
                 switch(m_compiler)
                 {
                 case Compiler::Gcc:
                     // Handled with asm() later
                     break;
                 case Compiler::Msvc:
```
bjorn3, line 2010: This is now unreachable due to the above if.

arlosi: The `if` was already there -- I changed `item.m_linkage.name` to `linkage_name` in the condition.

But you're right, it was unreachable, so I'll remove the check for Gcc in the `if`.

arlosi:
CI failure doesn't look related to my change.

PR #158, opened 2020-06-18
Mostly x32 fixes

nabijaczleweli:
mrustc disagrees on the size of one type (`union u_ZRG3c3std9panickingh0134Data2gG2c3std5panic16AssertUnwindSafe1gG3c17rustc_incremental7persist4loadh7closure16load_dep_g$d67c7ee6`) when building rustc 1.29.0, which, when ignored (or, indeed, `__attribute__((packed))` away), produces a rustc that panics with
```
thread 'main' panicked at 'assertion failed: ((probe).index() != idx_end)', rustc-1.29.0-src/src/libstd/collections/hash/map.rs:541:53
```

rustc may be terminally broken, but this mrustc does work for simple x32 executables, so I'm opening this PR either way, could be useful for someone smarter than me

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -306,7 +306,7 @@ $(OBJDIR)%.o: src/%.cpp
 $(OBJDIR)version.o: $(OBJDIR)%.o: src/%.cpp $(filter-out $(OBJDIR)version.o,$(OBJ)) Makefile
 	@mkdir -p $(dir $@)
 	@echo [CXX] -o $@
-	$V$(CXX) -o $@ -c $< $(CXXFLAGS) $(CPPFLAGS) -MMD -MP -MF $@.dep -D VERSION_GIT_FULLHASH=\"$(shell git show --pretty=%H -s)\" -D VERSION_GIT_BRANCH="\"$(shell git symbolic-ref -q --short HEAD || git describe --tags --exact-match)\"" -D VERSION_GIT_SHORTHASH=\"$(shell git show -s --pretty=%h)\" -D VERSION_BUILDTIME="\"$(shell date -uR)\"" -D VERSION_GIT_ISDIRTY=$(shell git diff-index --quiet HEAD; echo $$?)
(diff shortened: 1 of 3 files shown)


rustc-1.29.0-src.patch
```diff
@@ -83,3 +83,16 @@
 +pub fn check_for(x: Feature) -> bool {
 +    false
 +}
+
+# This test doesn't actually detect x32 properly
```
hvdijk, line 87: It actually does detect x32 properly, I think: unlike other environments, Rust requires target names to match its expectations exactly. While `x86_64-linux-gnux32` may be a valid triple in most environments, it is not a valid Rust triple. For most targets where mrustc uses a different triple name compared to what Rust uses, it works out okay because nothing looks at the exact triple name too closely, but if it doesn't work out here, it may be better to just match expectations and change the triple name.

Aside from that: I have adapted this PR to mrustc 0.9 except with the Rust-standard triple name, applied the fix for #167, quickly hacked together a fix for #168 (that I should clean up, reapply to the main branch, and submit as a PR), and with that mrustc for x32 seems to be in a very good state: it was able to successfully build rustc 1.29.0 for x32.

Disclaimer: building rustc for x32 requires patching even when mrustc isn't part of the picture; I had applied those patches to the version in mrustc as well to get a fair comparison. None of those patches were to the mrust parts of the code.

nabijaczleweli: Agree on the quad name, I must've been confused when writing that.

Did you arrive at something like [this](https://github.com/thepowersgang/mrustc/compare/master...nabijaczleweli:x32-v2)? It builds and the strings embedded in rustc look right, but I'd like confirmation from someone who has actually managed to use this before before I update the PR.

hvdijk: Yes, that looks like it's pretty much what I was using, I had a few minor differences, but it was basically just code style, nothing that should affect results. The `--no-show-signature` thing is unrelated to x32, but yes I have the same Git config so I needed that too. It might be good to split that into a separate PR.

nabijaczleweli: Great, thanks!

PR #159, opened 2020-06-20
Add initial support for ppc64

luke-jr:
Note, I couldn't get a complete build to work... Got as far as `(95/96) BUILDING cargo v0.30.0` though.

```
[RUSTC] -o output/prefix-s/lib/rustlib/powerpc64le-unknown-linux-gnu/lib/libcore.rlib
make: *** [Makefile:120: output/prefix-s/lib/rustlib/powerpc64le-unknown-linux-gnu/lib/libcore.rlib] Segmentation fault
```

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -44,6 +44,18 @@ const TargetArch ARCH_M68K = {
     { /*atomic(u8)=*/true, false, true, false,  true },
     TargetArch::Alignments(2, 2, 2, 2, 2, 2, 2)
 };
+const TargetArch ARCH_POWERPC64 = {
+    "powerpc64",
+    64, true,
+    { /*atomic(u8)=*/true, true, true, true,  true },
+    TargetArch::Alignments(2, 4, 8, 16, 4, 8, 8)
+};
+const TargetArch ARCH_POWERPC64LE = {
+    "powerpc64",
+    64, false,
+    { /*atomic(u8)=*/true, true, true, true,  true },
+    TargetArch::Alignments(2, 4, 8, 16, 4, 8, 8)
+};
 TargetSpec  g_target;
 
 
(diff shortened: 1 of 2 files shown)


r-burns:
It looks like the issue you ran into was not architecture-specific and has been fixed since you opened this. I was able to apply this patch on top of https://github.com/thepowersgang/mrustc/commit/db097f7849366ac560f6c4de4d1c7d0e7e768ab5 and bootstrap rustc 1.29 on powerpc64le-linux (using a workaround in https://github.com/thepowersgang/mrustc/issues/187).

thepowersgang:
1.39 is still the highest fully supported version.

r-burns:
FWIW, I built mrustc as a nix package, so it was using all of the deterministic build mechanisms enforced by nixpkgs stdenv such as fixed -frandom-seed and SOURCE_DATE_EPOCH. I don't know anything about how that random value is derived, so not sure if this is relevant.

thepowersgang:
It's not random, it's a stable hash of the crate ID and the (incremental-safe) definition ID.
I think the original issue I saw was due to not using enum niche optimisations where they were expected.

luke-jr:
Went to try master w/ 1.39 and realised I did already (ffb0961ad3a1125ca7db3b6a48991023bcb4d0d7 to be specific).

It failed pretty early with:

```
Process was terminated with signal 7
FAILING COMMAND:  /var/tmp/portage/dev-lang/rust-1.40.0_p20211027/work/rustc-1.40.0-src/mrustc/output/build_compiler_builtins-0_1_18_H22_run > /var/tmp/portage/dev-lang/rust-1.40.0_p20211027/work/rustc-1.40.0-src/mrustc/output/build_libc-0_2_62_Hd_run
```

thepowersgang:
`SIGBUS` - that's odd, does `make test` (which runs a basic "hello, world" work?

luke-jr:
This was during `RUSTC_VERSION=1.39.0 MRUSTC_TARGET_VER=1.39 OUTDIR_SUF=-1.39.0 make -f minicargo.mk LIBS`, so long before hello world would be applicable?

luke-jr:
FWIW, playing around, I tried 953b8a9e9c129e22338daf1a837852c9d20afc06 w/ 1.54.0 and got that same error

thepowersgang:
I guess you'll have to do some debugging (e.g. run the relevant program in `gdb`) to find the codegen error.

luke-jr:
Moved that topic to #194

PR #160, opened 2020-08-10
codegen_c: use command -v instead of which

Artoria2e5:
which needs to be installed. command is a posix builtin.


Artoria2e5:
What's the official position on whether to put a space after `for`, `if`, `while`, and `switch`? The code is quite inconsistent about it. Consider finding some time to make a `clang-format` rule.

thepowersgang:
Semi-official policy is no space between `for` and `(` - but a space between the `()` and the contents

thepowersgang:
Also - `clang-format` may have quite the time with the `TU_*` macros ;)

PR #172, opened 2021-04-12
i386-gnu target for GNU/Hurd

zamaudio:
This branch gets you most of the way to a complete port on GNU/Hurd.
There is a bug in rustc 1.29 sources that prevents this mrustc from compiling rustc on hurd.
(I believe the `popfd` and `pushfd` instructions are only valid when using `.intel_syntax` in GAS on i386)

We still need to port the libstd to os hurd target.

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -4183,7 +4183,7 @@ namespace {
 
             m_of << indent << "__asm__ ";
             if (is_volatile) m_of << "__volatile__";
-            m_of << "(\"" << (is_intel ? ".syntax intel; " : "");
+            m_of << "(\"" << (is_intel ? ".intel_syntax; " : "");
             // TODO: Use a more powerful parser
             for (auto it = e.tpl.begin(); it != e.tpl.end(); ++it)
             {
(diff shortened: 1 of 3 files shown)


thepowersgang:
Do either of 1.39 or 1.54 support hurd?
If they do, then I'm OK with merging support.

Otherwise, I'm not willing (or really able) to (formally) support a target that rustc doesn't.

bjorn3:
Hurd isn't supported on any rustc version afaik, not even master has it: https://github.com/rust-lang/rust/tree/master/compiler/rustc_target/src/spec You may want to propose hurd as tier 3 target for rustc. See https://rustc-dev-guide.rust-lang.org/building/new-target.html for how to add it and https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-3-target-policy for the tier 3 target policy you would have to follow.

averyterrel:
https://doc.rust-lang.org/stable/rustc/platform-support/hurd.html
Hurd is currently supported in stable

thepowersgang:
If that is part of 1.74, then that will be available once that branch is complete (see you in several months I guess)

Googulator:
Looks like it's brand new in 1.74 - https://doc.rust-lang.org/1.74.0/rustc/platform-support/hurd.html exists, https://doc.rust-lang.org/1.73.0/rustc/platform-support/hurd.html doesn't

thepowersgang:
This should be mergable after #340 340 then

thepowersgang:
1.74 support has been merged, mind fixing the merge conflicts and checking it it works?

PR #175, opened 2021-05-10
Fix $CC+triple lookup

dkm:
'-' is not a valid char for environment variable, use '_' instead.
eg. CC_x86_64_linux_gnu instead of invalid CC-x86_64-linux-gnu.

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -1111,12 +1111,14 @@ namespace {
             {
             case Compiler::Gcc:
                 // Pick the compiler
-                // - from `CC-${TRIPLE}` environment variable
-                // - from the $CC environment variable
-                // - `gcc-${TRIPLE}` (if available)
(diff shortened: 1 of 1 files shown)


dkm:
Sample test : 
```
$ CC_x86_64_linux_gnu=true ./bin/mrustc test.rs -o bob -L output/ 
...
(0.00 s) Trans Enumerate Cleanup: DONE
Trans Codegen: V V V
CC_x86_64_linux_gnu
Running command - "true" @"bob_cmd.txt"
(0.00 s) Trans Codegen: DONE
```
instead of : 
```
(0.00 s) Trans Enumerate Cleanup: DONE
Trans Codegen: V V V
CC_x86_64_linux_gnu
Running command - "x86_64-linux-gnu-gcc" @"bob_cmd.txt"
(0.00 s) Trans Codegen: DONE
```

I've also reordered the comment to match the actual code.

PR #176, opened 2021-05-11
Change the order in which GCC is searched

dkm:
Instead of using $CC-triple, triple-gcc, $CC, gcc,
Use $CC-triple, $CC, triple-gcc, gcc.

Makes it easier to override the compiler for every possible triple.


thepowersgang:
Not really sure which order is right

dkm:
Oh ?
My use case is to override the compiler without knowing what system is being used (using $CC-tripple is not possible).
Having the env vars checked first and only after checking what is available on the system makes it possible.
What are the use cases that would be broken by this change? Is it expected to have $CC defined and expect tripple-cc to take precedence ?

PR #179, opened 2021-05-26
Resolve Use - Try unprefixed paths in modules

Tarnyko:
When using imported crates in modules, there are
cases where paths are automatically module-prefixed
("crate::mymodule::core" instead of "core" e.g.);
if this fails, retry with no prefix.

### DIFF
--- a/src/resolve/use.cpp
+++ b/src/resolve/use.cpp
@@ -136,6 +136,8 @@ void Resolve_Use_Mod(const ::AST::Crate& crate, ::AST::Module& mod, ::AST::Path 
 {
     TRACE_FUNCTION_F("path = " << path);
 
+    ::AST::Path abs_path, abs_mod_path;
+
     for(auto& use_stmt : mod.m_items)
     {
         if( ! use_stmt->data.is_Use() )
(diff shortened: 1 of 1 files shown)


Tarnyko:
How to reproduce (mrustc 1.39.0 must be in "**$HOME/mrustc/**") :
```
$ wget http://www.tarnyko.net/repo/mrustc-bug_resolve_module.tar.bz2
$ tar xfvj mrustc-bug_resolve_module.tar.bz2
$ cd mrustc-bug_resolve_module/
$ ./build.sh
```
 …
_**error:0:Cannot find component 1 of crate::modtest::core::sync::atomic::AtomicUsize/*?*/ (_)**_

###  Detail :

- **test.rs** : 
```
extern crate core;
 […]
cfg_if! {
    if #[cfg(feature = "alloc")] {
        mod modtest;
    }
```
- **modtest.rs** :
`use core::sync::atomic::AtomicUsize;`

In the main source, we import the "core" crate; then there is a macro, and a module inside the macro.
In said module (modtest.rs) we "use" a member of the core crate imported in parent, but it is not found, because mrustc tries to prefix the path with the module itself ("_crate::modtest::core::sync::atomic::AtomicUsize_").

thepowersgang:
This seems more like a bug in the parser, not properly annotating the use path as being absolute.

Tarnyko:
As discussed, closing (not a proper fix, a good one would handle 2018/non-2018 modules & crates interfacing -in the compiler).

PR #183, opened 2021-07-19
Enable bootstrap of rustc 1.39.0 on Windows

arlosi:
* Adds bootstrapping scripts for 1.39.0 on Windows
* Increase stack size for mrustc to avoid stack overflow
* Add argument quoting for minicargo
* Minor fixes

### DIFF
--- a/vsproject/build_rustc_minicargo_139.cmd
+++ b/vsproject/build_rustc_minicargo_139.cmd
@@ -1,0 +1,41 @@
+@setlocal
+@call build_std_139.cmd
+@if %errorlevel% neq 0 exit /b %errorlevel%
+
+@IF "%LLVM_CONFIG%"=="" (
+    @REM Build LLVM
+    @mkdir %OUTDIR%llvm-build
+    pushd %OUTDIR%llvm-build
+    cmake "%~dp0..\rustc-%RUSTC_VERSION%-src\src\llvm-project\llvm" "-G" "Ninja" "-DLLVM_ENABLE_ASSERTIONS=OFF" ^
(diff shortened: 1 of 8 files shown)


vsproject/build_rustc_minicargo_139.cmd
```diff
@@ -0,0 +1,41 @@
+@setlocal
```
thepowersgang, line 1: Might be an idea to just `@echo off` to avoid prefixing every line (and turn it back on when it's time to run the actual build commands)

vsproject/mrustc.vcxproj
```diff
@@ -168,6 +168,7 @@
       <AdditionalDependencies>common_lib.lib;mrustc_lib.lib;%(AdditionalDependencies)</AdditionalDependencies>
       <AdditionalLibraryDirectories>$(OutDir)</AdditionalLibraryDirectories>
       <AdditionalOptions>/WHOLEARCHIVE:mrustc_lib.lib %(AdditionalOptions)</AdditionalOptions>
+      <StackReserveSize>8388608</StackReserveSize>
```
thepowersgang, line 171: Out of curiosity, what part needed the larger stack?

arlosi: IIRC it was while doing constant evaluation while compiling rustc (or one of its dependencies).

arlosi: I should also note that I only tested the full bootstrap when combined with my other PR to update to VS2019. I don't have the VS2015 compiler installed anymore. I expect it should work, but it's possible there may be some differences from the new compiler.

arlosi:
Parts of this change has already been merged via other PRs, and work has moved to 1.54. Closing.

PR #186, opened 2021-08-30
Fix gcc-11 on Ubuntu 21.04

chorman0773:
This patch fixes errors discovered locally using gcc-11 on ubuntu 21.04.

It adds a number of missing includes to various files, and readds expand/panic.o to fix a linker UNDEF (Expand_init_panic).


chorman0773:
I changed it to a draft because I am investigating some build errors for libcore with the patched code (may not be related to the patch, but I have no way of testing the unpatched version):
> > /home/chorm/mrustc/bin/mrustc rustc-1.39.0-src/src/libcore/lib.rs -o output-1.39.0/libcore.rlib --crate-name core --crate-type rlib -C emit-depfile=output-1.39.0/libcore.rlib.d --crate-tag 0_0_0 -g --cfg debug_assertions -O -L output-1.39.0 --edition 2018
rustc-1.39.0-src/src/libcore/slice/mod.rs:67: warn:0:Unexpected attribute allow_internal_unstable on associated item
rustc-1.39.0-src/src/libcore/str/mod.rs:2173: warn:0:Unexpected attribute allow_internal_unstable on associated item
MIR ERROR: <* as ::"core-0_0_0"::alloc::Alloc>::realloc_array BB7/TERM: Use of non-valid lvalue - _4
BUILD FAILED
make: *** [minicargo.mk:82: output-1.39.0/libstd.rlib] Error 1
Command exited with non-zero status 2
9.45user 0.18system 0:09.74elapsed 98%CPU (0avgtext+0avgdata 441780maxresident)k
16inputs+40outputs (0major+111709minor)pagefaults 0swaps

thepowersgang:
All fixed as of 474bec9cfd7862a20e7288cecd7fcf5e18648b9a

The non-valid value was an interaction around moving a vector that differed between MSVC and glibc.

PR #190, opened 2021-09-24
Fix macro_rules macros that match unions

arlosi:
`macro_rules!` macros that matched items failed to match `union`s because the parser did not account for the union having a name.

For example, the following macro failed to accept union type definitions:
```rust
macro_rules! __item {
    ($i:item) => {
        $i
    };
}
```


thepowersgang:
Ooh, damn - good spot, not sure how that didn't get hit earlier.

PR #192, opened 2021-10-18
Fix bootstrapping for macOS for 1.39.0

arlosi:
* Modify 1.39.0 patch to remove assertions about structure size due to alignment differences and workaround a linker issue searching for `CFMutableAttributedStringGetTypeID` by pulling in patch from `core-foundation`
* Fix incorrect signature in LLVM  `add_overflow` intrinsics
* Change the macOS target name to `x86_64-apple-darwin` to match Rust triple.

cc @woachk

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -1210,7 +1210,9 @@ namespace {
                     {
                         args.push_back(c);
                     }
+#if !defined(__APPLE__)
                     args.push_back("-Wl,--start-group");    // Group to avoid linking ordering
+#endif
                     //args.push_back("-Wl,--push-state");
                     for(auto l_d : libraries_and_dirs)
                     {
@@ -1240,7 +1242,9 @@ namespace {
                         }
                     }
                     //args.push_back("-Wl,--pop-state");
+#if !defined(__APPLE__)
                     args.push_back("-Wl,--end-group");    // Group to avoid linking ordering
+#endif
                     for( const auto& a : Target_GetCurSpec().m_backend_c.m_linker_opts )
                     {
                         args.push_back( a.c_str() );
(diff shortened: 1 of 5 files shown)


evanmiller:
Just the peanut gallery here, but can you explain why the `static_assert_size` calls have been removed?

thepowersgang:
Taking a serious look at this now (sorry it's taken so long).
- Could you update the patch file to make the size checks conditional on not using mrustc? `not(rust_compiler = "mrustc")`
- I'm still kinda curious as to why there was a null pointer in that deref chain
- Is `-darwin` always the correct target name? Or does `-macos` also valid (if it is also valid, then don't remove the target from the spec)

thepowersgang:
@arlosi Sorry, but a large amount of your changes have snuck in via other means.
Could you rebase this with just the remaining changes (and address the comments in https://github.com/thepowersgang/mrustc/pull/192#issuecomment-1001519202)

arlosi:
Sure, I can update the PR today.

* I'll add the conditional compilation for `not(rust_compiler = "mrustc")`
* Looks like the null deref question is resolved now.
* `-darwin` is always the correct name AFAIK. `-macos` isn't mentioned in the official rust compiler documentation.

src/trans/codegen_c.cpp
```diff
@@ -1210,7 +1210,9 @@ namespace {
                     {
                         args.push_back(c);
                     }
+#if defined(__APPLE__)
```
evanmiller, line 1213: Shouldn't this be `!defined`?

arlosi: Good catch. Fixed this locally when testing, but missed committing it.

src/trans/codegen_c.cpp
```diff
@@ -2478,10 +2482,10 @@ namespace {
...
                 }
-                // `fn llvm_addcarryx_u32(a: u8, b: u32, c: u32, d: *mut u8) -> u32`
+                // `fn llvm_addcarryx_u32(a: u8, b: u32, c: u32, d: *mut u8) -> u8`
                 else if( item.m_linkage.name == "llvm.x86.addcarryx.u32") {
-                    m_of << "\t*arg3 = __builtin_add_overflow" << msvc_suffix_u32 << "(arg1, arg2, &rv);\n";
-                    m_of << "\tif(*arg3) *arg3 |= __builtin_add_overflow" << msvc_suffix_u32 << "(rv, 1, &rv);\n";
+                    m_of << "\trv = __builtin_add_overflow" << msvc_suffix_u32 << "(arg1, arg2, arg3);\n";
+                    m_of << "\tif(arg0) rv |= __builtin_add_overflow" << msvc_suffix_u32 << "(*arg3, 1, arg3);\n";
                     m_of << "\treturn rv;\n";
                 }
```
thepowersgang, line 2490: Oof, that's a messy mistake. Took me quite a while to even be sure if the change was correct (it is... but was all ready to say it had an error).

arlosi:
I don't have access to the Mac hardware I used when I originally made this change, so I haven't re-validated the full bootstrap of mrustc -> 1.39.0 -> 1.39.0 -> 1.40.0 with the latest PR updates. It worked with the original version of the change.

I *have* used a Mac CI machine to test that mrustc can build rustc 1.39.0 with this updated change but I can't test further than that.

thepowersgang:
I believe that most of these changes were included in #241 - which has now been merged.

catap:
@thepowersgang if you willing to merge it, you may use version which is applied to last master and which I've used for my MacPorts port: https://github.com/catap/mrustc/commit/3d24d451b23104b58f8afc9d07d1cd341b296f5a

thepowersgang:
Wait, this is for 1.39's patch set, not 1.54's.
@arlosi Part of this has been applied, and some has been addressed. Could you update the patch set?

catap:
@thepowersgang my commit is for 1.39.0 also ;)

I can't move forward with 1.54.0.

thepowersgang:
Oh yeah, point still stands. This PR needs to be either updated against master, or closed.

arlosi:
Most of this change has already been merged via other PRs, and work has moved to 1.54. Closing.

PR #204, opened 2022-01-01
Use `$(DYLIB_EXT)` in one more place

evanmiller:
Inspired by #192


thepowersgang:
That was pretty quick :)

evanmiller:
Eager to get a New Year's build going here. The last bit of Makefile housekeeping are the `-linux` literals here if there's some way you'd like to support `-macos` also:

https://github.com/thepowersgang/mrustc/blob/8f52819e1f88d2fd7ecc7dd14c83df38b7fec7a7/run_rustc/Makefile#L118-L121

PR #206, opened 2022-01-02
Allow using system-installed LLVM

r-burns:
It's now possible to override MINICARGO and LLVM_CONFIG to use dependencies from outside the source tree. This is helpful for iterating on packaging, as the bootstrap can be split up into multiple packages.

### DIFF
--- a/minicargo.mk
+++ b/minicargo.mk
@@ -239,7 +239,7 @@ LLVM_CMAKE_OPTS += CMAKE_BUILD_TYPE=Release
 LLVM_CMAKE_OPTS += $(LLVM_CMAKE_OPTS_EXTRA)
 
 
-$(LLVM_CONFIG): $(RUSTCSRC)build/Makefile
+$(RUSTCSRC)build/bin/llvm-config: $(RUSTCSRC)build/Makefile
 	$Vcd $(RUSTCSRC)build && $(MAKE)
 
 $(RUSTCSRC)build/Makefile: $(RUSTCSRC)$(LLVM_DIR)/CMakeLists.txt


thepowersgang:
Has this been checked? `:=` in makefiles does an immediate assignment, maybe you wanted to use `?=`

r-burns:
I don't think immediate/lazy matters in this case since nothing is being expanded in the value, but happy to switch to ?= if that's more appropriate here

thepowersgang:
`?=` is "assign if not already assigned", which is more likely your intent.

thepowersgang:
Most of this is already applied through other PRs, mind taking a look and refining down to just the remaining changes?

r-burns:
Ah nice, thanks for the ping. Weird that the github diff view is not showing the conflict. Going to take a look at this, I think the LLVM_CONFIG change is still needed but everything else has been committed through other PRs.

minicargo.mk
```diff
@@ -239,7 +239,7 @@ LLVM_CMAKE_OPTS += CMAKE_BUILD_TYPE=Release
 LLVM_CMAKE_OPTS += $(LLVM_CMAKE_OPTS_EXTRA)
 
 
-$(LLVM_CONFIG): $(RUSTCSRC)build/Makefile
```
catap, line 242: `LLVM_CONFIG` is defined as:
```
LLVM_CONFIG := $(RUSTCSRC)build/bin/llvm-config
```

=> this isn't enough, isn't it?

r-burns: No, because I'm overriding LLVM_CONFIG on the command line. This rule does not specify how to build an arbitrary LLVM_CONFIG, it specifies how to build $(RUSTCSRC)/build/bin/llvm-config. When I specify some system LLVM I want this rule to not be triggered.

PR #211, opened 2022-01-05
Lex - include missing <limits> header

evanmiller:
Compilation fails with GCC 11 without this change.


thepowersgang:
Aah, the best part about C++ - You use an item and it happens to be transiently included by some random header.

PR #217, opened 2022-01-11
Move pattern into rule list

rustacean-xyz:
This should fix compilation on MSVC.


thepowersgang:
Thanks for spotting this, however your fix didn't work (see the failed appveyor build).

PR #221, opened 2022-01-24
Allow to use systems llvm

catap:


thepowersgang:
This seems like an iffy change - rustc is relatively tightly coupled to its version of llvm.
I don't want to have to support more variability in the build environment than I have to, and using an arbitrary LLVM seems like a recipe for extra work.

catap:
@thepowersgang the root cause of this PR is my attempt to bring rust to macOS 10.6.

To do so I'd love to use your project as bootstrap :)

Anyway, the challenge that LLVM/clang supports TLV only for macOS 10.7+ :)

MacPorts has a custom LLVM/clang with emulation TLV which works very nice, and I'm building `mrustc` with MacPorts LLVM.

catap:
I found much better way => let me close this one

catap:
Yep, I defently need it for my port file :)

catap:
Yep, I defently need it for my port file :)

catap:

thepowersgang:
You will want to remove/handle the `$(LLVM_CONFIG): $(RUSTCSRC)build/Makefile` line in `minicargo.mk` (so it doesn't apply when using the system LLVM)

catap:
After thinking a while I'd love to close this PR because it isn't good idea.

PR #222, opened 2022-01-24
Added `LINKFLAGS_EXTRA` and `CXXFLAGS_EXTRA`

catap:
This commit added `CXXFLAGS_EXTRA` at each `Makefile` and also introduce a brand new environment variable `LINKFLAGS_EXTRA` which can be used to adjust linking options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,6 +82,8 @@ ifneq ($(GPROF),)
   EXESUF := -gprof$(EXESUF)
 endif
 
+LINKFLAGS += $(LINKFLAGS_EXTRA)
+
 BIN := bin/mrustc$(EXESUF)
 
 OBJ := main.o version.o
--- a/bnf/Makefile
+++ b/bnf/Makefile
@@ -5,6 +5,10 @@
 .SUFFIXES:
 
 CXXFLAGS := -std=c++11 -x c++ -I . -g
+LINKFLAGS := -std=c++11 -g
+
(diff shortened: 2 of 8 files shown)


thepowersgang:
What is the goal/need for this change?

catap:
@thepowersgang to use custom `libcxx` which contains symbol `__cxa_thread_atexit` which is used by emulated TLV on macOS before 10.7

thepowersgang:
I'm assuming "TLV" is "Thread local values"
If there's special handling required for that version of OSX, maybe it should be integrated properly?

catap:
@thepowersgang I doubt that "proper way" is possible.

MacPorts implements an emulation of Thread Local Value storage for macOS before 10.7 via patch to its clang / llvm (which should be backported to rust's llvm) and an implementation such emulated TLV somewhere. MacPorts uses its libcxx for such ports.

My idea is build mrustc on macOS 10.6 with used macPorts environment and its patches.

Why? Because to real use macOS 10.6 you need fresh curl, openssl and tons of dependencies :)

catap:

thepowersgang:
I'd suggest also adding this to the other application makefiles (minicargo, dump_hirfile, ...)

catap:
@thepowersgang good idea. I also added a few missed `CXXFLAGS_EXTRA`.

Issue #223, opened 2022-01-25
Segmentation fault inside minicargo


catap:
I'm using 6f42b7415dc6c91ba41f8afb7d9e807e7c139c38 as local root.

An attmept to build `output-1.39.0/rustc` cause `Segmentation fault`. A make was run with `-j12`.

The report:
```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               minicargo [14544]
Path:                  /opt/local/var/macports/*/minicargo
Identifier:            minicargo
Version:               ???
Code Type:             X86-64 (Native)
Parent Process:        Exited process [14511]
Responsible:           Terminal [571]
User ID:               502

Date/Time:             2022-01-25 01:26:30.0500 +0100
OS Version:            macOS 12.1 (21C52)
Report Version:        12
Bridge OS Version:     6.1 (19P647)
Anonymous UUID:        F4C73193-A183-1E9B-1CEC-012963CBBDCA

Sleep/Wake UUID:       79CA752A-5561-4F47-A586-17BE96D23E02

Time Awake Since Boot: 190000 seconds
Time Since Wake:       33164 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [14544]

VM Region Info: 0 is not in any region.  Bytes before following region: 4453474304
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                      10972a000-10977e000    [  336K] r-x/r-x SM=COW  ...s/*/minicargo

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_platform.dylib      	    0x7ff81dcd66f2 _platform_strlen + 18
1   minicargo                     	       0x109740212 std::__1::char_traits<char>::length(char const*) + 8 (__string:372) [inlined]
[...]

PR #224, opened 2022-01-25
Fixed segmentation fault on huge paths

catap:
When path to a folder where mrustc huge the hardcoded buffer for `1024` char might not be enoug.

When it happened it corrupts memmory which leads to segfault.

I have no idea how to get `PATH_MAX` on windows => I've touched only not windows build.

### DIFF
--- a/tools/common/path.cpp
+++ b/tools/common/path.cpp
@@ -10,6 +10,7 @@
 # include <Windows.h>
 #else
 # include <unistd.h>    // getcwd/chdir
+# include <limits.h> // PATH_MAX
 #endif
 
 helpers::path::path(const char* s):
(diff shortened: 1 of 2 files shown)


catap:
Thus, it also may be a root cause for https://github.com/thepowersgang/mrustc/issues/223

catap:
Nope, it isn't :(

thepowersgang:
Both uses of these buffers include the maximum buffer length - they shouldn't be triggering a segmentation fault.
Do you have a backtrace that points to these faulting due to a too-small buffer?

(Side note: Looks like the windows version of the `getcwd` call doesn't check the return)

catap:
@thepowersgang segfault nope, but I have corrupted memory with very-very-very long paths which cause some random failrue.

I can try to find it in termnial scrolling history, if it survived.

catap:
@thepowersgang to be honest: I can't reproduce this error anymore, and it seems like false alarm.

If you not willing to merge it => let me now, and I'll close this PR.

thepowersgang:
It's a valid fix, I've just been drained by general life and checking the rest of your PRs.
Once CI passes I'll merge.

catap:
@thepowersgang I've migrated to universal header and now it should pass CI.

PR #231, opened 2022-01-28
Introduced `LLVM_CMAKE_OPTS_EXTRA`

catap:
Use cases: the way to suggest used python and libtool for cmake to build llvm on macOS 10.6.


catap:

PR #233, opened 2022-01-28
`run_rustc` respect `j` option

catap:


catap:

Makefile
```diff
@@ -142,10 +143,10 @@ PCHS := ast/ast.hpp
 OBJ := $(addprefix $(OBJDIR),$(OBJ))
 
 
-all: $(BIN)
+all: $(MRUSTC)
 
 clean:
-	$(RM) -r $(BIN) $(OBJ) bin/mrustc.a
+	$(RM) -r $(MRUSTC) $(OBJ) bin/mrustc.a
```
thepowersgang, line 149: Seems risky to delete the user-provided mrustc on clean.
Might want to keep `BIN` around and default `MRUSTC` to the same value.

Makefile
```diff
@@ -167,7 +168,9 @@ RUSTC_SRC_DL := $(RUSTCSRC)/dl-version
 MAKE_MINICARGO = $(MAKE) -f minicargo.mk RUSTC_VERSION=$(RUSTC_VERSION) RUSTC_CHANNEL=$(RUSTC_SRC_TY) OUTDIR_SUF=$(OUTDIR_SUF)
 
 
-output$(OUTDIR_SUF)/libstd.rlib: $(RUSTC_SRC_DL) $(BIN)
+$(MINICARGO): $(RUSTC_SRC_DL) $(MRUSTC)
```
thepowersgang, line 171: minicargo doesn't depend on these - It has its own source directory.

If you're going this far, might be an idea to move all of the "build rust code" logic to `minicargo.mk`

run_rustc/Makefile
```diff
@@ -109,19 +119,21 @@ all: $(BINDIR)rustc
...
-../output$(OUTDIR_SUF)/rustc:
-	$(MAKE) -C ../ output$(OUTDIR_SUF)/rustc -j 3
-../output$(OUTDIR_SUF)/cargo:
-	$(MAKE) -C ../ output$(OUTDIR_SUF)/cargo -j 3
+$(MINICARGO):
+	$(MAKE) -C ../ bin/minicargo$(EXESUF)
+$(MRUSTC_RUSTC):
+	$(MAKE) -C ../ output$(OUTDIR_SUF)/rustc
+$(MRUSTC_CARGO):
+	$(MAKE) -C ../ output$(OUTDIR_SUF)/cargo
```
thepowersgang, line 127: Probably worth keeping these using the default paths - So it errors out properly if you override `MRUSTC_RUSTC` to something non-exoistant

thepowersgang:
Side note - This PR doesn't seem to do anything about handling `-j` properly, can you rename it to something more fitting of the content?

Also - Please don't force push so often, I get a notification for every one with no indiciation of what changed.

catap:
@thepowersgang to make things simpler: I've split this PR into two. This one that matches the naming and https://github.com/thepowersgang/mrustc/pull/237

I also significant simplified it and avoid touching dependencies

Issue #234, opened 2022-01-30
Can't build `1.39.0` because `wasm32` arch is missed


catap:
`wasm32` is heavy used inside sources, and attempt to build it fails with error:
```
  EXCEPTION: Unknown cfg value `target_arch` (="wasm32")
```

thepowersgang:
This looks like it's coming from minicargo, can you provide more context for the error?
Minicargo has its own debug environment variable - `MINICARGO_DEBUG=Load Dependencies` might provide the context.

This error message should only show if mrustc failed to dump the `target_arch` variable - it's not because `wasm32` is unsupported.

catap:
@thepowersgang it was a bit tricky to obtain, but I'll try to reproduce it with suggested debug.

catap:
I can't reproduce an issue anymore.

Near the same moment I had a typo in `Cargo.toml` with line like `[target."cfg(not(any(target_os = \"windows\", target_os = \"wasm32\", target_os = \"macos\")))".dependencies.openssl]`. Maybe it was the root cause of this issue.

PR #235, opened 2022-01-30
Added `wasm32` target

catap:
It closes https://github.com/thepowersgang/mrustc/issues/234

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -59,6 +59,13 @@ const TargetArch ARCH_POWERPC64LE = {
     { /*atomic(u8)=*/true, true, true, true,  true },
     TargetArch::Alignments(2, 4, 8, 16, 4, 8, 8)
 };
+// see https://webassembly.org/docs/portability/
+const TargetArch ARCH_WASM32 = {
+    "wasm32",
+    32, false,
+    { /*atomic(u8)=*/true, true, true, true,  true },
+    TargetArch::Alignments(2, 4, 8, 8, 4, 8, 4)
+};
 TargetSpec  g_target;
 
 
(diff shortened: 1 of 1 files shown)


src/trans/target.cpp
```diff
@@ -59,6 +59,13 @@ const TargetArch ARCH_POWERPC64LE = {
     { /*atomic(u8)=*/true, true, true, true,  true },
     TargetArch::Alignments(2, 4, 8, 16, 4, 8, 8)
 };
+// see https://webassembly.org/docs/portability/
+const TargetArch ARCH_WASM32 = {
+    "wasm32",
+    32, false,
+    { /*atomic(u8)=*/true, true, true, false,  true },
```
bjorn3, line 66: Wasm32 supports 64bit atomics with the atomics extension. Without it doesn't support any atomics at all.

src/trans/target.cpp
```diff
@@ -59,6 +59,13 @@ const TargetArch ARCH_POWERPC64LE = {
...
+    TargetArch::Alignments(2, 4, 4, 4, 4, 4, 4)
```
bjorn3, line 67: This should be 2, 4, 8, 8, 4, 8, 4.

src/trans/target.cpp
catap: @bjorn3 thanks

src/trans/target.cpp
catap: @bjorn3 thanks

thepowersgang:
Unless you've tested that this 1. fixes the stated error, and 2. works (i.e. can compile working code), I'm not willing to merge it.

catap:
@thepowersgang after applying this patch I've overstep https://github.com/thepowersgang/mrustc/issues/234.

catap:
Regarding working code: I haven't tested it.

catap:

thepowersgang:
> This PR is required and used by [macports/macports-ports#13817](https://github.com/macports/macports-ports/pull/13817)

Citation needed. It's very unlikely that wasm is needed to compile for macos.

catap:
I can build without this PR. Seems that an issue was introduced by another issue.

Shall I close this PR for now or would you like to keep it?

PR #237, opened 2022-02-05
Introduce `MRUSTC`, `MINICARGO` and `LIBS_DIR`

catap:
Such environment variables is used to override used `mrustc`, `minicargo` and its libs


thepowersgang:
See comments on the last PR, might be a better idea to make the main makefile just directly defer to minicargo.mk for common operations (i.e. just for `test`), so there's no need to have the various rust compilation flags in it.

catap:
@thepowersgang may you add a link to the comment? :)

thepowersgang:
Huh, I thought I had suggesterd suggested that in one of your PRs, there's been so many and its been quite hot.

catap:
@thepowersgang to make things clear: my goal is bring `mrustc` into MacPorts to build the rust on macOS 10.6 :)

Right now I've split `mrustc` into two ports:
 - `mrustc` which contains `mrustc`, `minicargo` and libs;
 - `mrustc-rust` which contains rust and cargo which was compiled by `mrustc` from the first port.

The first port is buit by `-f minicargo.mk bin/mrustc bin/minicargo LIBS` and the second one as `-C run_rustc`.

This PR allows me to not re-compile everything when the second port is building.

catap:
@thepowersgang any chance to move forward with it? :)

thepowersgang:
I'd prefer to rewrite the interaction of the top-level `Makefile` and `minicargo.mk` first. They're too tightly coupled currently (which leads to issues when using `-j`)

catap:
@thepowersgang `-j` issue is much more deeply to be honest: at some point `minicargo` tries to compile the same file from different paths and `.c` files are huge => it may lead to case when file partially written or truncated when `cc` starts to compile it.

The right way is compile (both `build_` and not-`build_`) to some temporary folder and when it's done => move it to the right place.

It also protects from `Ctrl+C` which may broke compiling.

thepowersgang:
`minicargo` (if only one instance is running) should never try to compile the same file twice.
The problem happens if two instances of `minicargo` run at one time (using the same output directory and an overlapping dependency tree)

catap:
🤷‍♂️

but when I add `PARLEVEL=12` and run it under `-j12` it fails. Anyway, I feel that it also fails with `-j1` and `PARLEVEL=12` but I can't be sure, I'll run some test and comeback .

catap:
@thepowersgang after last huge session of debug I see why idea with `LIBS_DIR` is wrong.

Anyway, if I simplified it to contains only a few replacement with `MRUSTC` and `MINICARGO`... will you merge it?

thepowersgang:
As said above - I'd prefer to make `Makefile` be _just_ for building mrustc, which removes the need for any of the changes to that file.

Merging of this is on hold until that's done.

catap:
@thepowersgang I preferred when to not touch it because it quire fragile :) This patch suits well for me goal and I can ship it with port file that I'm making for last weeks.

catap:
@thepowersgang from another hand I guess I figured out that you mean.

This changes allows to simple run:
```
env RUSTC_VERSION=1.54.0 MRUSTC_TARGET_VER=1.54 OUTDIR_SUF=-1.54.0 make -C run_rustc all
```
and build everything from scratch on clean sources

PR #239, opened 2022-02-05
run_rustc: allowed to control `MINICARGO_FLAGS` and `CARGO_FLAGS`

catap:


catap:
Seems as very dangerous changes that makes things quite fragile. It is so simple to achive:
```
error: Package `XXX v0.0.0 (...)` does not have these features: `abc, def, ghk`
```

Better to close it.

PR #240, opened 2022-02-06
Suppress some warning

catap:


thepowersgang:
See above: MSVC doesn't like `_Pragma`, and gcc is very warning happy.

catap:
@thepowersgang I may drop the second commit if you would like the first one.

And wrapping all code by `#prgma ... #pragma` seems like a wrong idea.

Do you agree?

thepowersgang:
The fixes seem acceptable, it's just the pragma annotations that are difficult.

Issue #138, opened 2020-01-19
Fails to link cargo


dtolnay:
When running `make -f minicargo.mk output/cargo`, one of the commands mrustc 0.9 tries to run is:

```bash
[...]

PR #241, opened 2022-02-06
Bootstrapping for 1.54.0 for macOS 

catap:
Here, two small patches that allows me to build rust on macOS 12.1 as
```
env OPENSSL_DIR=/opt/local/libexec/openssl11 ./build-1.54.0.sh
env RUSTC_VERSION=1.54.0 MRUSTC_TARGET_VER=1.54 OUTDIR_SUF=-1.54.0 make -C run_rustc all
```

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -2478,10 +2478,10 @@ namespace {
                     m_of << "\tif(arg0) rv._0 |= __builtin_add_overflow" << msvc_suffix_u32 << "(rv._1, 1, &rv._1);\n";
                     m_of << "\treturn rv;\n";
                 }
-                // `fn llvm_addcarryx_u32(a: u8, b: u32, c: u32, d: *mut u8) -> u32`
+                // `fn llvm_addcarryx_u32(a: u8, b: u32, c: u32, d: *mut u8) -> u8`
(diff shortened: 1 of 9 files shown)


src/trans/codegen_c.cpp
```diff
@@ -1210,7 +1210,9 @@ namespace {
                     {
                         args.push_back(c);
                     }
+#if !defined(__APPLE__)
```
bjorn3, line 1213: This should probably get the target somehow rather than defining this based on the host.

catap: @bjorn3 probably, anyway, it seems as easy way to do.

Also here: https://github.com/thepowersgang/mrustc/issues/138 you may read near the same discussion.

catap: BTW: used Compiler is hardcoded inside the code :)

thepowersgang: Go ahead and remove this part of the changeset, I'm making it a part of `m_linker_opts`

catap: @thepowersgang here it is.

catap:
@thepowersgang let me rebase to confirm that it works

catap:
`./build-1.54.0.sh` works

PR #243, opened 2022-02-07
Dump loaded `lang_item`s from hir

catap:
Here a two places where it should be dump:
 - inside `dump_hirfile`;
 - inside `from_ast.cpp`.

### DIFF
--- a/src/hir/from_ast.cpp
+++ b/src/hir/from_ast.cpp
@@ -2232,6 +2232,7 @@ public:
             lang_item_path.first,
             HIR::SimplePath(g_crate_name, lang_item_path.second.nodes)
             ) );
+            DEBUG("Defined language item '" << lang_item_path.first << "' at " << lang_item_path.second);
     }
     rv.m_ext_crates_ordered = crate.m_extern_crates_ord;
     for(auto& ext_crate : crate.m_extern_crates)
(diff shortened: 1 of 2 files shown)


thepowersgang:
Good idea, although the wording of the messages needs a little tweaking - and the `dump_hirfile` output should include the path as well as the name.

catap:
@thepowersgang any suggestion for name tweaking ?

src/hir/from_ast.cpp
```diff
@@ -2242,6 +2243,7 @@ ::HIR::CratePtr LowerHIR_FromAST(::AST::Crate crate)
             const auto& name = lang.first;
             const auto& path = lang.second;
             auto irv = rv.m_lang_items.insert( ::std::make_pair(name, path) );
+            DEBUG("Load language item '" << lang.first << "' at " << lang.second << " for " << ext_crate.first);
```
thepowersgang, line 2246: "from" not "for" would be more suitable.

tools/dump_hirfile/main.cpp
```diff
@@ -86,6 +86,11 @@ void Dumper::dump_crate(const char* name, const ::HIR::Crate& crate) const
     }
 #endif
 
+    for(const auto& i : crate.m_lang_items)
+    {
+        ::std::cout << "// lang "<< i.first << ::std::endl;
```
thepowersgang, line 91: Add the path too

catap:
@thepowersgang here it is

Issue #236, opened 2022-02-01
Tons of warnings on macOS


catap:
Each time when I've tried to rebuild mrustc on macOS I have a lot of warnings like this.
```
[CXX] -o .obj/main.o
warning: anonymous non-C-compatible type given name for linkage purposes by typedef declaration; add a tag name here [-Wnon-c-typedef-for-linkage]
src/hir/path.hpp:43:9: note: type is not C-compatible due to this default member initializer
        unsigned index = ~0u;
        ^~~~~~~~~~~~~~
src/include/tagged_union.hpp:215:17: note: expanded from macro 'TAGGED_UNION_EX'
    TU_TYPEDEFS _variants/*
                ^~~~~~~~~
src/include/tagged_union.hpp:191:73: note: expanded from macro 'TU_TYPEDEFS'
#define TU_TYPEDEFS(...)     TU_EXP1( TU_GMX(__VA_ARGS__)(TU_TYPEDEF   ,__VA_ARGS__) )
                                                                        ^~~~~~~~~~~
src/include/tagged_union.hpp:36:40: note: expanded from macro 'TU_DISP4'
#define TU_DISP4(n, v, v2, v3, v4)   n v n v2 n v3 n v4
                                       ^
src/include/tagged_union.hpp:165:42: note: expanded from macro 'TU_TYPEDEF'
#define TU_TYPEDEF(name, ...)    typedef __VA_ARGS__ TU_DATANAME(name);/*
                                         ^~~~~~~~~~~
src/include/tagged_union.hpp:23:21: note: expanded from macro 'TU_EXP1'
#define TU_EXP1(x)  x
                    ^
src/hir/path.hpp:41:1: note: type is given name 'Data_Infer' for linkage purposes by this typedef declaration
TAGGED_UNION_EX(ConstGeneric, (), Infer, (
^
src/include/tagged_union.hpp:215:5: note: expanded from macro 'TAGGED_UNION_EX'
    TU_TYPEDEFS _variants/*
    ^
src/include/tagged_union.hpp:191:59: note: expanded from macro 'TU_TYPEDEFS'
#define TU_TYPEDEFS(...)     TU_EXP1( TU_GMX(__VA_ARGS__)(TU_TYPEDEF   ,__VA_ARGS__) )
                                                          ^
warning: anonymous non-C-compatible type given name for linkage purposes by typedef declaration; add a tag name here [-Wnon-c-typedef-for-linkage]
[...]

PR #244, opened 2022-02-08
Fixed some warnings from `clang` (ref #236)

catap:
Here two fixes:
 - the correct suppression for clang's nitpicking
 - minor fixes for `typeid` side effect

### DIFF
--- a/src/include/tagged_union.hpp
+++ b/src/include/tagged_union.hpp
@@ -208,7 +208,18 @@
  * ```
  */
 #define TAGGED_UNION(_name, _def, ...)  TU_EXP1( TAGGED_UNION_EX(_name, (), _def, (TU_EXP(__VA_ARGS__)), (), (), ()) )
+#if defined(__clang__)
 #define TAGGED_UNION_EX(_name, _inherit, _def, _variants, _extra_move, _extra_assign, _extra) \
+_Pragma("clang diagnostic push"); \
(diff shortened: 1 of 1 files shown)


thepowersgang:
I'm not sure about the usage of `#pragma`, may make a decision this week.

catap:
Just remember about this one.

thepowersgang:
The `typeid` changes seem a little clunky, (you're making a new variable but only using it once).
Still not 100% on the pragma, could do with comments about why it has to be in that header.

thepowersgang:
Is it feasible to add that option to the command-line?

catap:
@thepowersgang what shall we do with this one?

thepowersgang:
As is probably obvious from the lack of a merge, I'm not very happy with this changeset as it currently applies.

- The pragma hackery could go in its own PR (with a comment on it explaining the reasoning) and it'd probably be OK.
- The `typeid` changes should have more effort applied (to give that new variable more usage)... or just suppress the error in the makefiles.

catap:
@thepowersgang I've removed `typeid` commit and keep only pragma.

thepowersgang:
Cool.

The `typeid` change could probably be re-done in a different PR by making it a method of the `HIR::ExprNode` type, that would be cleaner than using `typeid` directly every time.

Issue #214, opened 2022-01-06
LIBS build error with GCC 11 (Undefined symbol `___atomic_compare_exchange_16`)


evanmiller:
Getting closer but the following error appears with GCC 11 / Mac OS X 10.6:

```
(2/26) BUILDING cmake v0.1.38
> bin/mrustc rustc-1.39.0-src/vendor/cmake/src/lib.rs -o output-1.39.0/libcmake-0_1_38.rlib --crate-name cmake --crate-type rlib -C emit-depfile=output-1.39.0/libcmake-0_1_38.rlib.d --crate-tag 0_1_38 -g --cfg debug_assertions -O -L output-1.39.0 --extern cc=output-1.39.0/libcc-1_0_35.rlib
> bin/mrustc rustc-1.39.0-src/vendor/libc/build.rs --crate-name build --crate-type bin -o output-1.39.0/build_libc-0_2_62_Hd_run -L output-1.39.0 -g --cfg feature=rustc-dep-of-std --cfg feature=align --cfg feature=rustc-std-workspace-core
Undefined symbols for architecture x86_64:
  "___atomic_compare_exchange_16", referenced from:
      _ZRG4cE9core0_0_09core_arch6x86_6410cmpxchg16b_D0g in libcore.rlib.o
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
```

Maybe `mrustc` needs to add another GCC flag or libatomic linkage?

bjorn3:
This should use the cmpxchg16b instruction. Does passing -mcpu=core2 to gcc work? That is what rustc passes to llvm on macOS. https://github.com/rust-lang/rust/blob/f1ce0e6a00593493a12e0e3662119786c761f375/compiler/rustc_target/src/spec/x86_64_apple_darwin.rs#L6

evanmiller:
-mcpu=core2 works on a small test program.

evanmiller:
The very first Intel Macs were 32-bit "Core Solo" and "Core Duo". Poking around EveryMac, it appears that the first 64-bit Macs were in fact `core2`. So I think it's safe to add to the `x86_64` target.

evanmiller:
This error still appears when attempting to compile with GCC 11 / 10.6.8 / x86_64.

thepowersgang:
With the above PR applied? can you confirm that `-march` is being passed?

evanmiller:
[...]

PR #245, opened 2022-02-08
Enforce `-mtune=core2` and `-mtune=yonah` on macOS

catap:
I've used the same cpu which is used by Rust: `core2` for x86_64[1] and `yonah` for i686[2]. I also used `mtine` instead of `mcpu` because it was deprecated at gcc since version 5[3].

It closes https://github.com/thepowersgang/mrustc/issues/214

I also added missed `i686-apple-darwin` target.

Footnotes:
[1]  https://github.com/rust-lang/rust/blob/1.54.0/compiler/rustc_target/src/spec/x86_64_apple_darwin.rs
[2]  https://github.com/rust-lang/rust/blob/1.54.0/compiler/rustc_target/src/spec/i686_apple_darwin.rs
[3]  https://github.com/gcc-mirror/gcc/commit/9d913bbf3fc996874649168d7d144a642012ac9b

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -541,11 +541,21 @@ namespace
                 ARCH_X86_64
                 };
         }
+        else if(target_name == "i686-apple-darwin")
+        {
+            // NOTE: OSX uses Mach-O binaries, which don't fully support the defaults used for GNU targets
+            // The first 32bit Intel Mac was Core Solo aka yonah. It allows to use `-march=yonah` like Rust.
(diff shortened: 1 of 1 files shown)


bjorn3:
`-mtune` merely optimizes for better performance on a specific cpu while still allowing it to run on older systems, this means it doesn't enable target features supported by the specified cpu like xchg16b support. You should probably use `-march` instead.

catap:
@bjorn3 very good catch. Let me fix it.

PR #247, opened 2022-02-09
Missed dependency to build run-rustc

catap:
`$(LIBDIR_2)libtest.rlib` is used `$(BINDIR)cargo` but doesn't depend on it.


thepowersgang:
I assume you found this by trying to do massive `-j` calls?

catap:
and using `-C run_rustc all` as the end point to build everything.

PR #249, opened 2022-02-10
Obtain crash from remote build

catap:
This PR is depends on https://github.com/thepowersgang/mrustc/pull/248 and it was found by this one :)

Why? Let assume that I run CI as github action and at some point bootstrap fails like this `Process was terminated with signal 13`.

I haven't got access to a disk and can't read `_dbg.txt` file, nor see the crash report.


thepowersgang:
This seems a little too specialized to include. Ideally you'd be debugging issues outside of CI

PR #254, opened 2022-02-13
A few trivial fixes that allows build standalone_miri on macOS

catap:
The fail:
```
[CXX] miri_extern.cpp
miri_extern.cpp:630:20: error: no matching function for call to 'clock_gettime'
        int rv_i = clock_gettime(clk_id, reinterpret_cast<struct timespec*>(tp_vr.data_ptr_mut()));
                   ^~~~~~~~~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/time.h:178:5: note: candidate function not viable: no known conversion from 'unsigned int' to 'clockid_t' for 1st argument
int clock_gettime(clockid_t __clock_id, struct timespec *__tp);
    ^
miri_extern.cpp:749:27: error: use of undeclared identifier 'memrchr'; did you mean 'memchr'?
        const void* ret = memrchr(ptr, c, n);
                          ^~~~~~~
                          memchr
```

### DIFF
--- a/tools/standalone_miri/miri_extern.cpp
+++ b/tools/standalone_miri/miri_extern.cpp
@@ -28,7 +28,7 @@
 #undef DEBUG
 
 
-#ifdef _WIN32
+#if _WIN32 || __APPLE__
 const char* memrchr(const void* p, int c, size_t s) {
     const char* p2 = reinterpret_cast<const char*>(p);
     while( s > 0 )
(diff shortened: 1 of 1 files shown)


thepowersgang:
Did you test this? You might want to split that `#if-#else-#endif` up so OSX still gets `write` and `sysconf` defined?

catap:
@thepowersgang `make -C tools/standalone_miri` works fine on local root b97168299e1b38e52fe8266a727f5b7be5aac383 (this PR) on my laptop with macOS 12.

Anyway, I haven't tried macOS 10.6 yet. I can do it but in a few hours.

catap:
@thepowersgang seems that macOS has both `sysconf` and `write` at least at [10.2.8](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX10.2.8.sdk/usr/include/unistd.h#L133).

Issue #256, opened 2022-02-16
Inconsistent setting of OUTDIR_SUF


r-burns:
OUTDIR_SUF is set inconsistently: in minicargo.mk it is always set to -$(RUSTC_VERSION), but in run_rustc it is empty if rustc is considered the default version (currently 1.29.0).

https://github.com/thepowersgang/mrustc/blob/87b7481c54b7cb4379cdb5126a2cf7cf9a4f9b83/minicargo.mk#L15-L21

https://github.com/thepowersgang/mrustc/blob/87b7481c54b7cb4379cdb5126a2cf7cf9a4f9b83/run_rustc/Makefile#L28-L32

This causes missing target errors during run_rustc bootstrap of rustc 1.29:

```
make[1]: *** No rule to make target 'output/rustc'.  Stop.
make[1]: Leaving directory '/build/source'
make: *** [Makefile:122: ../output/rustc] Error 2
make: *** Waiting for unfinished jobs....
make[1]: *** No rule to make target 'output/cargo'.  Stop.
make[1]: Leaving directory '/build/source'
make: *** [Makefile:124: ../output/cargo] Error 2
make: Leaving directory '/build/source/run_rustc'
```

PR #257, opened 2022-02-16
Synchronize `OUTDIR_SUF` for default version

catap:
`OUTDIR_SUF` produces different value when default version is used between `minicargo.mk` and `run_rustc`.

Closes: https://github.com/thepowersgang/mrustc/issues/256

Cc: @r-burns

### DIFF
--- a/minicargo.mk
+++ b/minicargo.mk
@@ -12,8 +12,12 @@ endif
 # ----- INPUTS ------
 # -------------------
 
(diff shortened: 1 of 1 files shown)


r-burns:
To be honest I think it would be simpler to just always add the RUSTC_VERSION suffix to the outdir, but either way works for me. It's simple to override the OUTDIR as desired anyway. Thanks for the fix!

PR #259, opened 2022-02-16
Use `F_DUPFD_CLOEXEC` only on macOS 10.7+

catap:
`F_DUPFD_CLOEXEC` was introduced in macOS 10.7 like this:
```
#if __DARWIN_C_LEVEL >= 200809L
#define	F_DUPFD_CLOEXEC		67	/* mark the dup with FD_CLOEXEC */
#endif
```


thepowersgang:
This removes `F_DUPFD_CLOEXEC` for all other platforms.

catap:
@thepowersgang indeed. Here should be much better.

thepowersgang:
Now I know you're not testing things :) - should be `||`

catap:
@thepowersgang 🤦, usually I did but not this one because to test it isn't something easy. Have I missed something?

thepowersgang:
The second version of this commit would have also never exposed that option (even on OSX)

PR #260, opened 2022-02-17
Fix -Wmaybe-uninitialized warnings in move constructors

r-burns:
With GCC 9, x.m_ptr was triggering -Wmaybe-uninitialized.
This seems to silence the warnings without changing the behavior.

(This is purely cosmetic, so feel free to close if you prefer the existing style. This warning can pop up in other spurious contexts so you may prefer to silence it entirely.)

### DIFF
--- a/src/hir/type_ref.hpp
+++ b/src/hir/type_ref.hpp
@@ -7,6 +7,7 @@
 */
 #pragma once
 
+#include <utility>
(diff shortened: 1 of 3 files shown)


thepowersgang:
Seems strange that this would be triggering a warning, seems like pretty simple logic

PR #278, opened 2022-07-24
Removed side effect from typeid call (ref #236)

catap:
Can build it now with `-Werror=potentially-evaluated-expression`

### DIFF
--- a/src/ast/expr.cpp
+++ b/src/ast/expr.cpp
@@ -21,6 +21,10 @@ ExprNodeP::ExprNodeP(std::unique_ptr<ExprNode> node)
     : m_ptr(node.release())
 {
 }
+const char* ExprNodeP::type_name() const
+{
+    return typeid(*m_ptr).name();
+}
 
 Expr::Expr(ExprNodeP node):
     m_node(node.release())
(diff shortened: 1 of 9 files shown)


thepowersgang:
Why `const std::type_info& type_info() const;` instead of `const char* type_name() const;`

catap:
@thepowersgang indeed

thepowersgang:
Seems to be missing the deref - You used `typeid(m_ptr)` instead of `typeid(*m_ptr)`

catap:
@thepowersgang this way it fails like:
```
src/ast/expr_ptr.hpp:46:44: error: 'typeid' of incomplete type 'AST::ExprNode'
    const char* type_name() const { return typeid(*m_ptr).name(); }
```

thepowersgang:
The implementation needs to be in `expr_ptr.cpp` (or wherever the destructor implementation is)

catap:
@thepowersgang thanks!

thepowersgang:
Have you tested that the debug output is still good/useful?

catap:
@thepowersgang all => nope, some => yes; but as you pointed with `*m_ptr` I defently did miss something.

catap:
@thepowersgang I've tested it via https://github.com/thepowersgang/mrustc/pull/279 and it produces good/useful debug output.

PR #279, opened 2022-07-24
`MRUSTC_DEBUG=*` enables all debug options

catap:


thepowersgang:
An interesting change, but questionable utility. The debugging output is MASSIVE, and you will always want t filter it to just a specific pass.

catap:
@thepowersgang I've needed it a few time to capture debug output to the log and grep by it.

PR #281, opened 2022-07-30
Fix includes

Lucius-Q-User:


thepowersgang:
Could you provide context as to why this change was needed? (Platform, compiler, ...)

src/ast/dump.cpp
```diff
@@ -11,6 +11,7 @@
 #include <main_bindings.hpp>
 #include <hir/hir.hpp>  // ABI_RUST - TODO: Move elsewhere?
 #include <fstream>
+#include <limits>
```
catap, line 14:
```suggestion
#include <limits> // std::numeric_limits
```

catap:
@thepowersgang I'd like to confirm that I have the same error with `gcc-mp-12 (MacPorts gcc12 12.1.0_6+stdlib_flag) 12.1.0`

Issue #284, opened 2022-08-07
Fail to build `LIBS` on macOS 10.7


catap:
Local root `42b54861f26253a2ab2c71bcfbe74ee2418db293` which is failed as:
```
(0/8) BUILDING cfg_if from cfg-if v0.1.10
> /opt/local/var/macports/build/_opt_bblocal_var_buildworker_ports_build_ports_lang_mrustc/mrustc/work/mrustc-42b54861f26253a2ab2c71bcfbe74ee2418db293/bin/mrustc rustc-1.54.0-src/vendor/cfg-if-0.1.10/src/lib.rs -o output-1.54.0/libcfg_if-0_1_10.rlib --crate-name cfg_if --crate-type rlib -C emit-depfile=output-1.54.0/libcfg_if-0_1_10.rlib.d --crate-tag 0_1_10 --cfg debug_assertions -O -L output-1.54.0 --edition 2018
Assertion failed: (t == ":"), function load_depfile, file build.cpp, line 817.
make: *** [output-1.54.0/libpanic_unwind.rlib] Abort trap: 6
```

The full log is available here: https://build.macports.org/builders/ports-10.7_x86_64-builder/builds/108738/steps/install-port/logs/stdio

catap:
Wired. It produces `output-1.54.0/libcfg_if-0_1_10.rlib.d` with content:
```
output-1.54.0/libcfg_if-0_1_10.rlib: rustc-1.54.0-src/vendor/cfg-if-0.1.10/src/lib.rs
output-1.54.0/libcfg_if-0_1_10.rlib: output-1.54.0/libcore.rlib
```
without the last `\n`.

PR #286, opened 2022-08-08
Compatibility with macOS 10.7

catap:
Let assume that `::std::istream` marks itself as reached `EOF` after reading the last character from file which is `':'` in case of minicargo's `build.cpp`.

This leads to missing `':'` because the next call of `get_token_int()` returns `""` because `ifp.eof()` is `true`, instead of returning `":"` as it should.

This missed `':'` leads to assert that dependency file aren't well formatted on some wired system.

The same happened with `Lexer::getc_byte` when `m_istream.get()` returns not `EOF` but marks stream as reached `EOF` that means the next read returns `EOF`.

Thus, this fixes: https://github.com/thepowersgang/mrustc/issues/284

### DIFF
--- a/src/parse/lex.cpp
+++ b/src/parse/lex.cpp
@@ -1102,7 +1102,7 @@ uint32_t Lexer::parseEscape(char enclosing)
 char Lexer::getc_byte()
 {
     int rv = m_istream.get();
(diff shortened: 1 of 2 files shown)


catap:
@thepowersgang and I'd like to remind about this one

PR #287, opened 2022-08-10
Introducing `MRUSTC_LIB` env variable (ref #214)

catap:
The idea is simple: we need some way to put `-latomic` to `gcc` call. Let add a dedicated variable which allows to add any library which may be needed.

It fixes: https://github.com/thepowersgang/mrustc/pull/214

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -893,6 +893,11 @@ ProgramParams::ProgramParams(int argc, char *argv[])
         this->lib_search_dirs.push_back(a);
     }
 
+    if( const auto* a = getenv("MRUSTC_LIBD") )
+    {
+        this->libraries.push_back(a);
+    }
+
     // Hacky command-line parsing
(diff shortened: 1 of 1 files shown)


catap:
@thepowersgang I'd like to remind about this one

thepowersgang:
I don't feel this should be a mrustc environment variable, maybe a minicargo one - but if this is required for a target, it should be added to the target spec.

catap:
@thepowersgang but at some point of bootstrap minicargo is replaced to cargo, and such `-latomic` still needed for gcc.

Or I've missed something?

thepowersgang:
minicargo is used whenever mrustc is used.
The `PROXY_MRUSTC` variable in `run_rustc` is only ever set to the mrustc-built rustc

catap:
After thinking a while it seems like a bad approach. Let me close it.

PR #290, opened 2022-09-12
Windows build - fix compilation errors on VS2022

arlosi:
A few minor changes necessary to make `mrustc` build with VS2022.

Does not actually change any of the projects to build with VS2022.

### DIFF
--- a/src/include/tagged_union.hpp
+++ b/src/include/tagged_union.hpp
@@ -18,6 +18,7 @@
 //#include "cpp_unpack.h"
 #include <cassert>
 #include <string>
+#include <stdexcept>
 
 #define TU_FIRST(a, ...)    a
 #define TU_EXP1(x)  x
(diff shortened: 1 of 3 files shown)


thepowersgang:
Appveyor build errored with `..\..\src\mir\mir_builder.cpp(741): error C2280: 'SplitArm::SplitArm(void) noexcept': attempting to reference a deleted function [C:\projects\mrustc\vsproject\mrustc_lib\mrustc_lib.vcxproj]`

arlosi:
Trying a fix. If it doesn't work I'll install VS2015 locally to get it passing.

arlosi:
I couldn't find a way to make it work on both 2015 and 2022, so I had to make it conditional on `_MSC_VER`

thepowersgang:
Hmm... I'm slightly hesitant to add guarded changes like that.
What's the goal of targetting VS2022? (Laziness will keep me on 2015 until there's a good reason to upgrade)

arlosi:
No specific reason other than VS2015 [ended mainstream support on Oct 13, 2020](https://learn.microsoft.com/en-us/lifecycle/products/visual-studio-2015)

I've removed the conditional compilation parts of this PR. Let's at least get those parts in.

PR #292, opened 2022-10-01
Never use section on aarch64 (#277)

### LINKED ISSUE #277: Fails to bootstrap rust-1.54.0 on aarch64 / macOS 12
I'm running on a local root 2c4f89f87692f27246d1bc1e42c8d43df8514a71 `./build-1.54.0.sh` which is failed as:
```
> /Users/catap/src/mrustc/bin/mrustc -Z print-cfgs
(0/13) BUILDING core v0.0.0
> /Users/catap/src/mrustc/bin/mrustc rustc-1.54.0-src/library/core/src/lib.rs -o output-1.54.0/libcore.rlib --crate-name core --crate-type rlib -C emit-depfile=output-1.54.0/libcore.rlib.d --crate-tag 0_0_0 --cfg debug_assertions -O -L output-1.54.0 --edition 2018
:0: BUG:src/mir/from_hir.cpp:524: TODO: visit - asm! const
FAILING COMMAND:  /Users/catap/src/mrustc/bin/mrustc rustc-1.54.0-src/library/core/src/lib.rs -o output-1.54.0/libcore.rlib --crate-name core --crate-type rlib -C emit-depfile=output-1.54.0/libcore.rlib.d --crate-tag 0_0_0 --cfg debug_assertions -O -L output-1.54.0 --edition 2018
BUILD FAILED
make: *** [output-1.54.0/libstd.rlib] Error 1

```

catap:
Closes #277

### DIFF
--- a/rustc-1.54.0-src.patch
+++ b/rustc-1.54.0-src.patch
@@ -115,3 +115,137 @@
 +#[cfg(any(miri, rust_compiler = "mrustc", not(all(feature = "simd", any(target_arch = "x86_64")))))]
  use self::generic as arch;
 
+#
+# Backport which is required to support arm64 on macOS 12
+# See: https://github.com/alexcrichton/curl-rust/commit/0aea09c428b9bc2bcf46da0fc33959fe3f03c74a
+#
+diff --git vendor/curl/src/lib.rs vendor/curl/src/lib.rs
+index 9f2e50ea9b..50eaba742b 100644
+--- vendor/curl/src/lib.rs
++++ vendor/curl/src/lib.rs
+@@ -82,6 +82,9 @@ pub mod easy;
+ pub mod multi;
+ mod panic;
+ 
++#[cfg(test)]
++static INITIALIZED: std::sync::atomic::AtomicBool = std::sync::atomic::AtomicBool::new(false);
++
+ /// Initializes the underlying libcurl library.
+ ///
+ /// The underlying libcurl library must be initialized before use, and must be
+@@ -102,46 +105,62 @@ pub fn init() {
+     /// Used to prevent concurrent or duplicate initialization.
+     static INIT: Once = Once::new();
+ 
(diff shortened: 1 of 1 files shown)


catap:
@thepowersgang I not sure about condition. Maybe invert it and keep section attribute only for `x86` and `x86_64` is better idea.

catap:
@thepowersgang just a reminder about this one

catap:
@thepowersgang ping here!

thepowersgang:
I'm not sure hard-disabling the use of section annotations is the correct solution to that linked issue.
Wouldn't adding patch for that particular bad annotation be a better solution?

catap:
@thepowersgang I haven't found the place in sources which triggers such behaviour. Anyway, section isn't supported at aarch64 => seems logical to not use it.

thepowersgang:
If that's the case - then include that in the commit and in a source comment

catap:
@thepowersgang seems I found much better solution.

I've started to find any proof that aarch64 doesn't support section and haven't found one. In fact I found that it should work.

=> I've started digging as you suggested and found an issue in rust-curl which is used. This issue was fixed in upstream. So, I've backported it to rust-curl which is used for 1.54.0 :)

catap:
Thus, you may see that original issue https://github.com/alexcrichton/curl-rust/issues/417 is near the same that I had at the end of https://github.com/thepowersgang/mrustc/issues/277:
```
  = note: ld: reference to symbol (which has not been assigned an address) __ZN4curl4init9INIT_CTOR17h97cc33cf050cb462E in '__ZN4curl4init17ha644d831c2a57f65E' from /Users/cormac/git/tryout/libcurl-monterey/target/debug/deps/libcurl-0f9cbb7dde66dd88.rlib(curl-0f9cbb7dde66dd88.curl.0b6dcf6e-cgu.2.rcgu.o) for architecture arm64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Issue #300, opened 2023-04-10
Support PowerPC: mrustc itself builds fine, rust tries to build for x86 and fails


barracuda156:
What is needed to support building for PowerPC? Currently `mrustc` builds, but `rust` tries to build for x86:
```
test -e bin/mrustc
/usr/bin/make -C tools/minicargo/
make[1]: Nothing to be done for `all'.
test -e bin/minicargo
bin/minicargo rustc-1.54.0-src/library/std --vendor-dir rustc-1.54.0-src/vendor --script-overrides script-overrides/stable-1.54.0-macos/ --output-dir output-1.54.0/  --manifest-overrides rustc-1.54.0-overrides.toml
> /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_mrustc/mrustc/work/mrustc-11a9c01119988e30ba40bab4ee5f60753c0c4b61/bin/mrustc -Z print-cfgs
(0/13) BUILDING core v0.0.0
> /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_mrustc/mrustc/work/mrustc-11a9c01119988e30ba40bab4ee5f60753c0c4b61/bin/mrustc rustc-1.54.0-src/library/core/src/lib.rs -o output-1.54.0/libcore.rlib --crate-name core --crate-type rlib -C emit-depfile=output-1.54.0/libcore.rlib.d --crate-tag 0_0_0 --cfg debug_assertions -O -L output-1.54.0 --edition 2018
rustc-1.54.0-src/library/core/src/../../stdarch/crates/core_arch/src/x86/avx2.rs:520: error:0:Type mismatch between [u32; 8] and [u32; 576460752303423488] - sizes differ
rustc-1.54.0-src/library/core/src/../../stdarch/crates/core_arch/src/x86/avx2.rs:520: note: From here
Process was terminated with signal 6
FAILING COMMAND:  /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_mrustc/mrustc/work/mrustc-11a9c01119988e30ba40bab4ee5f60753c0c4b61/bin/mrustc rustc-1.54.0-src/library/core/src/lib.rs -o output-1.54.0/libcore.rlib --crate-name core --crate-type rlib -C emit-depfile=output-1.54.0/libcore.rlib.d --crate-tag 0_0_0 --cfg debug_assertions -O -L output-1.54.0 --edition 2018
BUILD FAILED
[...]

PR #302, opened 2023-04-13
Fix targets for Darwin PowerPC

barracuda156:
See: https://github.com/thepowersgang/mrustc/issues/300#issuecomment-1506104456

@thepowersgang Could you please check this?

P. S. As I mention in the topic, Darwin on ppc32 _may_ need 4-byte bools, since Apple ABI defines that: http://personal.denison.edu/~bressoud/cs281-s07/MacOSXLowLevelABI.pdf
Not sure if that should apply to Rust code though.

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -59,6 +59,12 @@ const TargetArch ARCH_POWERPC64 = {
     { /*atomic(u8)=*/true, true, true, true,  true },
     TargetArch::Alignments(2, 4, 8, 16, 4, 8, 8)
 };
+const TargetArch ARCH_POWERPC32 = {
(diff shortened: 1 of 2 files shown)


thepowersgang:
Seems reasonable, although - do you want to wait until the PPC support is working?

barracuda156:
@thepowersgang I think this can be merged, as long as we are okay with targets (IMO they are correct), since `mrustc` itself builds and runs, what is broken is upstream `rust`, which can be fixed separately. Having ppc targets fixed may also be helpful in a case someone else will join the endeavor to fix `rust` for ppc, FWIW.
But of course this is up to you to decide.

PR #305, opened 2023-06-24
Add missing includes for GCC >= 13, Clang

orowith2os:
Just one of some patches needed to ensure Rust builds with newer GCC and Clang, the other patches involve modifying the rust source code itself.


orowith2os:
Oh, this includes the -j stuff I tried but failed with. That shouldn't be there.

PR #314, opened 2023-09-03
Lifetime Annotations and Inference

thepowersgang:
Although the branch is called `borrowcheck` - this is just the first half of a borrow checker.

Includes:
- Storage and use of lifetime parameters through all compilation phases.
- Lifetime elision rules
- Lifetime inference - hecking of function inputs and outputs, required for closures to have lifetime annotations
- Lots of improvements to promotion of borrows of constants.


thepowersgang:
Looks like an important patch was clobbered in the merge from master, waiting for full regression test with fixed patch.

thepowersgang:
Full regression run passes!

Issue #324, opened 2023-11-22
Unable to build against musl


lrvick:
Unable to bootstrap 1.54 from musl or glibc based linux distributions. 

## musl (alpine)

Repeated with master, and v0.10.1
```
docker run -it alpine
apk add git make gcc bash curl patch g++ zlib-dev
git clone https://github.com/thepowersgang/mrustc
cd mrustc
make -C run_rustc RUSTC_VERSION=1.54.0 RUSTC_TARGET=x86_64-unknown-linux-musl
```

```
[...]
[CXX] -o bin/mrustc
objcopy --only-keep-debug bin/mrustc bin/mrustc.debug
objcopy --add-gnu-debuglink=bin/mrustc.debug bin/mrustc
strip bin/mrustc
make[2]: Leaving directory '/mrustc'
test -e bin/mrustc
make -C tools/minicargo/
make[2]: Entering directory '/mrustc/tools/minicargo'
[CXX] main.cpp
[CXX] manifest.cpp
[CXX] repository.cpp
[CXX] cfg.cpp
[CXX] build.cpp
[CXX] jobs.cpp
[CXX] file_timestamp.cpp
[CXX] os.cpp
[CXX] -o ../../bin/minicargo
objcopy --only-keep-debug ../../bin/minicargo ../../bin/minicargo.debug
objcopy --add-gnu-debuglink=../../bin/minicargo.debug ../../bin/minicargo
strip ../../bin/minicargo
make[2]: Leaving directory '/mrustc/tools/minicargo'
test -e bin/minicargo
tar -xf rustc-1.54.0-src.tar.gz
cd rustc-1.54.0-src/ && patch -p0 < ../rustc-1.54.0-src.patch;
patching file compiler/rustc_ast/src/ast.rs
patching file compiler/rustc_hir/src/hir.rs
patching file compiler/rustc_middle/src/mir/interpret/error.rs
patching file compiler/rustc_middle/src/mir/mod.rs
patching file compiler/rustc_middle/src/thir.rs
patching file compiler/rustc_mir/src/interpret/operand.rs
patching file compiler/rustc_mir/src/interpret/place.rs
patching file library/stdarch/crates/std_detect/src/detect/mod.rs
patching file vendor/crc32fast/src/specialized/mod.rs
patching file vendor/curl/src/lib.rs
patching file vendor/ppv-lite86/src/lib.rs
patching file src/llvm-project/llvm/include/llvm/Support/Signals.h
[...]

PR #325, opened 2023-11-28
musl support

lrvick:
Simple patch to add musl support.

There is no reliable way to detect musl I am aware of, but we can detect glibc which defines _USE_GNU, then we fall back to musl.

This successfully builds on alpine. 

LMK if this approach works for you, and I can at least take a look at CI.

Linking #324

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -425,6 +425,13 @@ namespace
                 ARCH_X86_64
                 };
         }
+        else if(target_name == "x86_64-linux-musl" || target_name == "x86_64-unknown-linux-musl")
+        {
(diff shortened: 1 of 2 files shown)


thepowersgang:
Well, it passes CI.
I assume that you've tested it with your own musl setup?

lrvick:
Confirmed the above was related to gcc13 removing automatic includes of stdint, which was fixed in later versions of rust. We will need to backport a patch to make rust 1.54 compile with gcc13 but a non issue for this PR.

I have now confirmed builds both in alpine and my own from-scratch musl-based distro with musl 1.2.4 and GCC12.

PR #327, opened 2024-01-13
Use $OBJCOPY and $STRIP instead of hardcoding

lanodan:


tools/minicargo/Makefile
```diff
@@ -41,9 +41,9 @@ $(BIN): $(OBJS) ../../bin/common_lib.a
 ifeq ($(OS),Windows_NT)
 else ifeq ($(shell uname -s || echo not),Darwin)
 else
-	objcopy --only-keep-debug $(BIN) $(BIN).debug
-	objcopy --add-gnu-debuglink=$(BIN).debug $(BIN)
-	strip $(BIN)
+	$(OBJCOPY) --only-keep-debug $(BIN) $(BIN).debug
```
thepowersgang, line 44: You forgot to add the `OBJCOPY ?= objcopy` to this file

lanodan:
>You forgot to add the `OBJCOPY ?= objcopy` to this file

Right, I badly assumed it somehow wouldn't be needed in this file due to $CXX being used but absent, which seems a bit weird.

PR #344, opened 2024-12-09
Fix target-name for powerpc64le.

Millak:
* tools/common/target_detect.h: Fix typo in powerpc64le-* triplets.


thepowersgang:
Huh, looks like df243f74b3267b20430b818277c7e9e1de9951ad broke it.

PR #373, opened 2025-11-08
Actually fix the build on darwin powerpc

barracuda156:
I will add a couple of commits soon, until then draft, but:

mrustc-2


thepowersgang:
The build image selected for windows has been removed, so those are broken.

PR #407, opened 2026-05-19
minicargo: write --help text to stdout, keep usage errors on stderr

rautyrauty:
This is necessary so that `minicargo --help | less` and similar pipelines behave the way users expect — the help output is what was asked for, not a diagnostic, and belongs on stdout. cargo, rustc and GNU tools all follow this split: explicit `--help` to stdout, usage errors (wrong flags, missing args) to stderr.

usage() is reused on the error path, so it now takes an ostream& and defaults to std::cerr; help() passes std::cout explicitly. No change for the error-path callers.

### DIFF
--- a/tools/minicargo/main.cpp
+++ b/tools/minicargo/main.cpp
@@ -64,7 +64,7 @@ struct ProgramOptions
     ::std::vector<::std::string>    features;
 
     int parse(int argc, const char* argv[]);
-    void usage() const;
+    void usage(::std::ostream& os) const;
     void help() const;
 };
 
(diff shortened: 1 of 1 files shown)


tools/minicargo/main.cpp
```diff
@@ -64,7 +64,7 @@ struct ProgramOptions
     ::std::vector<::std::string>    features;
 
     int parse(int argc, const char* argv[]);
-    void usage() const;
+    void usage(::std::ostream& os = ::std::cerr) const;
```
thepowersgang: Probably remove the default and explicitly pass `::std::cerr` when called

rautyrauty:
Done in 9a4a534b — dropped the default and passed `::std::cerr` explicitly from the error-path caller in `ProgramOptions::parse`. Force-pushed; the rest of the diff is unchanged.

PR #408, opened 2026-05-19
minicargo.mk: skip download/extract when rustc source dir exists

rautyrauty:
Previously, even if the rustc-X.Y.Z-src/ tree was already populated by hand (distro packagers extracting from .tar.zst, a git checkout, etc.), make would still try to curl rustc-X.Y.Z-src.tar.gz from static.rust-lang.org and then re-extract it. This made offline / sandboxed builds (ALT Linux hasher, Nix, Guix, ...) require forging empty backdated tarballs and stamp files just to trick make into skipping the steps.

Detect a pre-existing source directory at parse time with $(wildcard ...) and, in that case, replace the extract recipe with a no-op that just touches the stamp. The download recipe is dropped from this branch entirely since nothing depends on it anymore. The normal download-driven flow is preserved verbatim when the directory is absent.

The patch step (rustc-X.Y.Z-src.patch -> dl-version stamp) is left alone: if the packager has already applied the patch, they can touch the dl-version stamp themselves; otherwise it runs as before. This matches the existing stamp-driven model.

### DIFF
--- a/minicargo.mk
+++ b/minicargo.mk
@@ -217,7 +217,17 @@ bin/testrunner$(EXESUF):
 #
 # rustc (with std/cargo) source download
 #
+# If $(RUSTCSRC) (the extracted source directory) already exists, skip both the
+# download and extract steps -- this lets distro packagers / offline builds
+# pre-populate the tree from any source (.tar.zst, git checkout, ...) without
+# having to forge stamps. The patch step still runs unless its stamp is present.
 RUSTC_SRC_TARBALL := rustc-$(RUSTC_VERSION)-src.tar.gz
+ifneq ($(wildcard $(RUSTCSRC)),)
+# Pre-extracted source tree detected: no tarball needed, just touch the stamp.
+rustc-$(RUSTC_VERSION)-src/extracted:
+	@echo "[SKIP] extract (using pre-extracted $(RUSTCSRC))"
+	@touch $@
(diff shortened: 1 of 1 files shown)


thepowersgang:
Odd... I would expect that as long as `extracted` is newer than the tarball, then the existing logic should still work?

rautyrauty:
The timestamp check never runs because the tarball file is missing.

```make
$(RUSTC_SRC_TARBALL):
	curl ...

rustc-$(RUSTC_VERSION)-src/extracted: $(RUSTC_SRC_TARBALL)
```

With no `.tar.gz` on disk, make treats the prereq as needing to be built and runs the curl rule unconditionally — "newer than" only applies once both files exist. Touching `extracted` doesn't help; the packager would have to forge an empty backdated `.tar.gz` too.

thepowersgang:
Creating an empty tarball then creating the `extracted` file should be enough to avoid needing to add a workaround in the makefile

rautyrauty:
Fair — that's exactly the workaround we use today. It just means each downstream packager has to materialise an empty backdated binary file purely to fool the makefile, which trips up tools that audit SOURCES (rpmlint, sbom generators, reproducibility checks). This PR was an attempt to lift that out of every spec file.

Happy to close if you'd rather keep the rule unchanged.

PR #411, opened 2026-05-23
hir_expand/static_borrow: Handle Defer and array-literal elements in lifted statics

rautyrauty:
Two fixes to the Static Borrow Constants pass in
`src/hir_expand/static_borrow_constants.cpp`, both surfaced while bootstrapping
rustc 1.90.0 libcore for `aarch64-unknown-linux-gnu`.

## 1. Catch `HIR::Defer` from the extracted-static const-eval

The evaluator throws `Defer` when a dependency of the constant being evaluated
is not resolved yet, but the two SBC call sites of `evaluate_constant` had no
`try`/`catch`, so a deferred evaluation propagated to `std::terminate`.

Handle it the way `Constant Evaluate` already does: leave the new static's
value as not-yet-generated and let a later pass retry. This required promoting
the previously anonymous-namespace `Defer` into `namespace HIR` (a new
declaration in `src/hir_conv/constant_evaluation.hpp`) so out-of-module passes
can catch it.

## 2. Visit the element of a sized-array literal

`extract_node`'s visitor overrode `visit(HIR::ExprNode_ArraySized&)` in order
to remap a const-generic array *size*, but never chained to
`HIR::ExprVisitorDef::visit(node)`. The repeated element expression
(`node.m_val`) was therefore never visited at all.

Consequences for `[val; N]` when `val` mentions a const-generic:

* `is_generic` stayed clear, so `extract_node` wiped `params_def` and
  `constr_params` and emitted the lifted item as concrete;
* the `ExprNode_ConstParam` inside `val` never had its binding remapped into
  the new item's parameter space;
* `visit_type` was never applied to the element's result type.

Trans Monomorph then asserts:

```
BUG: ASSERT FAIL: src/hir_typeck/common.cpp:721:val.idx() < p->m_values.size():
  Value param N/*M:0*/ out of range for (max 0)
```

on `::"core-0_0_0"::core_arch::arm_shared::neon::generated::lifted#915`, whose
MIR is a bare `_0 = Constant(N/*M:0*/)`. The source is stdarch's aarch64 neon
intrinsics, e.g.

```rust
pub fn vqshlu_n_s8<const N: i32>(a: int8x8_t) -> uint8x8_t {
    static_assert_uimm_bits!(N, 3);
    unsafe { _vqshlu_n_s8(a, const { int8x8_t([N as i8; 8]) }) }
}
```

The fix is to call the base visitor first; `ExprVisitorDef` deliberately does
not descend into the array *size*, which is why the existing size handling
stays as it is.

## Testing

* `make -f minicargo.mk LIBS RUSTC_VERSION=1.90.0 MRUSTC_TARGET=aarch64-unknown-linux-gnu`
  — libcore/liballoc/libstd now get through every HIR and MIR pass (the run
  still stops at the cross C compile, which is a host-toolchain limitation,
  not an mrustc one).
* `make -f minicargo.mk LIBS RUSTC_VERSION=1.90.0` (host x86_64) — full
  standard library builds.
* `make -f minicargo.mk local_tests RUSTC_VERSION=1.90.0` — 30 passed,
  0 failed.

Note: with fix 2 in place, plain upstream + fix 2 no longer needs fix 1 to
build the 1.90.0 aarch64 library set. Fix 1 is kept as the defensive change it
is — `Defer` escaping to `std::terminate` is never the right outcome.

### DIFF
--- a/src/hir_expand/static_borrow_constants.cpp
+++ b/src/hir_expand/static_borrow_constants.cpp
@@ -910,7 +910,23 @@ namespace static_borrow_constants {
             } v(resolve, monomorph);
             node->visit(v);
             v.visit_type(node->m_res_type);
-            if( !v.is_generic ) {
+            // `is_generic` is set by the targeted visit overrides above
+            // (ExprNode_ConstParam, ExprNode_ArraySized) plus visit_type /
+            // visit_path_params. Those reliably cover type- and lifetime-
+            // generic references, but a const-generic value param can still
+            // leak through paths not yet enumerated -- e.g. ARM Neon's
+            // `core_arch::arm_shared::neon::generated` lifted const blocks
+            // hit `_ = Constant(N/*M:0*/)` in MIR with the surrounding
+            // method's `const N: i32` never remapped, then assert in
+            // MonomorphiserPP::get_value because the lifted static was
+            // wiped to a concrete (empty params) item. Preserve the params
+            // whenever the source context carries any const-generic value
+            // param: the only cost is a slightly later (Trans Monomorph)
+            // evaluation instead of an SBC-time concrete one.
+            bool has_value_generics =
+                   !m_resolve.impl_generics().m_values.empty()
+                || !m_resolve.item_generics().m_values.empty();
+            if( !v.is_generic && !has_value_generics ) {
                 params_def = HIR::GenericParams();
                 constr_params = HIR::PathParams();
                 DEBUG("Concrete static");
--- a/src/hir_expand/static_borrow_constants.cpp
+++ b/src/hir_expand/static_borrow_constants.cpp
@@ -1159,8 +1159,16 @@ namespace static_borrow_constants {
                     if( !new_static.m_params.is_generic() )
                     {
                         new_static.m_value.m_state->stage = ::HIR::ExprState::Stage::Sbc;
-                        new_static.m_value_res = ::HIR::Evaluator(sp, m_crate, nvs).evaluate_constant( new_static_pair.path, new_static.m_value, new_static.m_type.clone());
-                        new_static.m_value_generated = true;
+                        try {
+                            new_static.m_value_res = ::HIR::Evaluator(sp, m_crate, nvs).evaluate_constant( new_static_pair.path, new_static.m_value, new_static.m_type.clone());
+                            new_static.m_value_generated = true;
+                        }
+                        catch(const ::HIR::Defer&) {
+                            // A dependency is not fully resolved yet; leave the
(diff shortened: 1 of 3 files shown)


src/hir_expand/static_borrow_constants.cpp
```diff
@@ -910,7 +910,23 @@ namespace static_borrow_constants {
...
+            // generic references, but a const-generic value param can still
+            // leak through paths not yet enumerated -- e.g. ARM Neon's
+            // `core_arch::arm_shared::neon::generated` lifted const blocks
+            // hit `_ = Constant(N/*M:0*/)` in MIR with the surrounding
+            // method's `const N: i32` never remapped, then assert in
+            // MonomorphiserPP::get_value because the lifted static was
+            // wiped to a concrete (empty params) item. Preserve the params
+            // whenever the source context carries any const-generic value
+            // param: the only cost is a slightly later (Trans Monomorph)
+            // evaluation instead of an SBC-time concrete one.
```
thepowersgang: This is a really verbose comment, and points to a different error. The visitor above should be able to detect that const-generic and flag the generic-ness.
How is that const-generic still getting though without being flagged?

thepowersgang (changes requested):
The catching is good, but the logic for disabling lifting when there are value-generics in scope is questionable.

src/hir_expand/static_borrow_constants.cpp
rautyrauty: You were right that it pointed at a different error - dropped that commit for a one-liner.

The visitor's `ExprNode_ArraySized` override never chained to
`HIR::ExprVisitorDef::visit(node)`, so the element of `[val; N]` was never
visited: `is_generic` stayed clear, the `ExprNode_ConstParam` inside never had
its binding remapped, and `visit_type` never ran on its result type.

The input is `const { int8x8_t([N as i8; 8]) }` in stdarch's aarch64 neon
intrinsics (rustc 1.90.0 libcore); the lifted item's MIR is a bare
`_0 = Constant(N/*M:0*/)`, and Trans Monomorph asserts:

```
BUG: ASSERT FAIL: src/hir_typeck/common.cpp:721:val.idx() < p->m_values.size(): Value param N/*M:0*/ out of range for (max 0)
```

Calling the base visitor first fixes it; `ExprVisitorDef` deliberately skips
the array *size*, so the existing `m_size` handling is unchanged.

With this, the 1.90.0 aarch64 library set builds without the `Defer` catch
too. Kept it since `Defer` reaching `std::terminate` is never right, but happy
to drop it.

thepowersgang (approved):

PR #412, opened 2026-05-23
trans/codegen_c: Collapse Rust asm! double-dollar escape for GCC inline asm

rautyrauty:
Rust `asm!` templates are LLVM asm templates: `$N` is an operand reference and
`$$` is the escape for a literal `$`. mrustc lowers them through the GCC
inline-asm protocol, where operands are `%`-prefixed and `$` has no escape, so
the doubled `$` survives into the assembly text.

With the `.intel_syntax noprefix` wrapper mrustc emits for x86, GAS then
evaluates `"$$0x80"` as a non-8-bit expression and rejects the `int`
instruction. This breaks rustix-1.0.8 on i586 — its i686 syscall wrappers use
`"int $$0x80"`.

## Change

The un-escaping lives in `operator<<(::std::ostream&, const FmtGccAsm&)`, next
to the existing `%`/`{`/`}`/`|` handling, rather than in a separate
pre-processing pass over the template. `FmtGccAsm` gains an `is_intel` flag:

* AT&T blocks: `$$` -> `$` (the immediate prefix is wanted);
* Intel blocks: `$$` -> nothing (GAS has no use for the AT&T immediate prefix
  and rejects it).

The "is this an intel-syntax block" test was already computed twice per
assembly block; it is now a local reused for the prologue, the epilogue and
the formatter.

This also applies the un-escaping to `global_asm!`, which shares `FmtGccAsm`
and is subject to exactly the same LLVM escaping rule.

## Testing

`bin/mrustc --target i586-unknown-linux-gnu` on a small `no_core` program
containing `asm!("int $$0x80", in("eax") 1usize, in("ebx") code)`:

```c
/* default (intel) */
__asm__ __volatile__(".intel_syntax noprefix; int 0x80;\n .att_syntax; " : ...);
/* options(att_syntax) */
__asm__ __volatile__("int $0x80;\n " : ...);
```

`make -f minicargo.mk local_tests RUSTC_VERSION=1.90.0` — 30 passed, 0 failed.

### DIFF
--- a/src/trans/codegen_c.cpp
+++ b/src/trans/codegen_c.cpp
@@ -5618,11 +5618,33 @@ namespace {
                 if( (Target_GetCurSpec().m_arch.m_name == "x86" || Target_GetCurSpec().m_arch.m_name == "x86_64") && !se.options.att_syntax )
                     m_of << ".intel_syntax noprefix; ";
                 bool escape_percent = true || !inputs.empty() || !outputs.empty();
+                // Rust's `asm!` templates are passed verbatim to LLVM, where `$$`
+                // is the escape for a literal `$` (LLVM IR uses `$N` for operand
+                // references). The GCC inline-asm backend has no such escape, and
+                // a stray `$` survives into the GAS input as e.g. `int $0x80` —
+                // which GAS rejects in `.intel_syntax noprefix` mode with
+                // "operand type mismatch". Collapse `$$` here: drop it entirely
+                // for Intel-syntax blocks (so `int $$0x80` -> `int 0x80`), and
+                // halve it for AT&T blocks (so the immediate prefix survives).
+                auto xlat_dollars = [&](const std::string& s) {
+                    std::string r;
+                    r.reserve(s.size());
+                    for(size_t i = 0; i < s.size(); ++i) {
+                        if(s[i] == '$' && i+1 < s.size() && s[i+1] == '$') {
+                            if(se.options.att_syntax) r += '$';
+                            i++;
+                        }
+                        else {
+                            r += s[i];
+                        }
+                    }
+                    return r;
+                };
                 for(const auto& l : se.lines)
                 {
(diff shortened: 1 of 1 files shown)


src/trans/codegen_c.cpp
```diff
@@ -5618,11 +5618,33 @@ namespace {
...
+                        if(s[i] == '$' && i+1 < s.size() && s[i+1] == '$') {
+                            if(se.options.att_syntax) r += '$';
+                            i++;
+                        }
+                        else {
+                            r += s[i];
+                        }
+                    }
+                    return r;
+                };
```
thepowersgang: This translation should be in `operator<<(::std::ostream& os, const FmtGccAsm& x)` instead of as pre-processing pass.

thepowersgang (changes requested):

src/trans/codegen_c.cpp
rautyrauty: Moved into `operator<<(::std::ostream&, const FmtGccAsm&)` as a `case '$'`
alongside the existing `%`/`{`/`}`/`|` handling; the pre-processing lambda is
gone. `FmtGccAsm` gains an `is_intel` field for it. The "is this an
intel-syntax block" test was already computed twice per assembly block, so it
is now a local shared with the prologue, the epilogue and the formatter.

Note that this also applies the un-escaping to `global_asm!`, which uses the
same formatter and is subject to the same LLVM escaping rule.

Verified with `--target i586-unknown-linux-gnu` on a small `no_core` program
containing `asm!("int $$0x80", in("eax") 1usize, in("ebx") code)`:

```c
/* default (intel) */
__asm__ __volatile__(".intel_syntax noprefix; int 0x80;\n .att_syntax; " : ...);
/* options(att_syntax) */
__asm__ __volatile__("int $0x80;\n " : ...);
```

thepowersgang (approved):

PR #415, opened 2026-07-25
toml: Support nested arrays, inline tables, and full escape set

danifunker:
The minimal TOML parser aborted the whole manifest on constructs that appear in real-world Cargo.toml packaging metadata (which minicargo never consults):

- Arrays previously accepted only string elements. Array-of-array and array-of-inline-table values (e.g. `[package.metadata.deb].assets` / `[package.metadata.generate-rpm].assets`) threw "Unexpected token in array value position". Now integers and bools are accepted, and nested `[...]` / `{...}` groups are consumed (balanced) and discarded via a new `TomlFile::skip_composite_value()`.

- `handle_escape` only knew `\"`, `\\`, `\n`. Added `\t \r \b \f`, the `\uXXXX` / `\UXXXXXXXX` forms (hex consumed, placeholder stored), and the multi-line-basic-string line-ending backslash (trims trailing whitespace through the next non-whitespace char). `m_line` is threaded in so the line counter stays correct across trimmed newlines.

Lets minicargo parse manifests that reuse a shared engine crate whose parent Cargo.toml carries deb/rpm/aur packaging sections, without editing that file.

### DIFF
--- a/tools/common/toml.cpp
+++ b/tools/common/toml.cpp
@@ -215,12 +215,30 @@ TomlKeyValue TomlFile::get_next_value()
             if( t.m_type == TomlToken::Type::SquareClose )
                 break;
 
-            // TODO: Recursively parse a value
-            // TODO: OR, support other value types
             switch(t.m_type)
             {
             case TomlToken::Type::String:
                 rv.value.m_sub_values.push_back(TomlValue { t.as_string() });
+                break;
+            case TomlToken::Type::Integer:
+                rv.value.m_sub_values.push_back(TomlValue { t.m_intval });
+                break;
+            case TomlToken::Type::Ident:
+                if( t.m_data == "true" || t.m_data == "false" ) {
+                    rv.value.m_sub_values.push_back(TomlValue { t.m_data == "true" });
+                }
+                else {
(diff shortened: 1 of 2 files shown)


tools/common/toml.cpp
```diff
@@ -215,13 +215,31 @@ TomlKeyValue TomlFile::get_next_value()
...
+                    throw ::std::runtime_error(::format(m_lexer, ": Unexpected identifier in array value position - ", t));
+                }
+                break;
+            // Nested array or inline table: consume it (balanced) and discard.
+            // minicargo only reads flat string/scalar arrays (features, paths,
+            // ...); the nested forms appear solely in `[package.metadata.*]`
+            // packaging sections that minicargo never consults, so skipping
+            // keeps the parse moving instead of aborting the whole manifest.
+            case TomlToken::Type::SquareOpen:
+            case TomlToken::Type::BraceOpen:
```
thepowersgang, line 242: Maybe print a warning, so this is caught if it ends up mattering.

danifunker: Good idea, should be updated in this PR now.

PR #417, opened 2026-07-31
Assorted fixes found building a large crate graph

danifunker:
Independent of any one target; each was hit building rusty-backup's dependency tree and none is specific to the architecture that surfaced it.

parse:
- Accept an interpolated item in a trait body. `Parse_Impl_Item` handled the shape but a trait body had no case at all, so `$item:item` in a trait failed with "Unexpected token TOK_INTERPOLATED_ITEM".
- Accept an interpolated associated const in an impl block.
- Transfer attributes parsed in front of an interpolated impl item onto it. They are parsed before the fragment is seen, so they were dropped - turning a `#[cfg]` that should remove the item into one that does nothing.
- Accept a non-numeric literal suffix. rustc's lexer takes any identifier as a suffix and only rejects it on evaluation, so a suffixed literal that only ever appears in a macro matcher is legal; the suffix is emitted as a following identifier token, which the matcher and the invocation both see identically.

proc macros:
- Send trait definitions to the plugin. `visit_item` had no Trait case, so an attribute macro on a `trait` aborted the compiler mid-conversation.
- Accept `#![crate_type = "proc-macro"]`; `Type::ProcMacro` was reachable only from `--crate-type`.
- Implement string-literal escapes in the plugin-side lexer.

typeck / const-eval:
- Revisit an ambiguous `_CallPath` instead of aborting, so a call whose callee is not yet resolved (an unresolved const-generic, say) defers rather than killing the compile.
- Pass an out-of-range lifetime through instead of asserting. Lifetimes are erased before codegen, so an unresolved one is inert - `get_type`/`get_value` still assert, because an unresolved type or const is not.
- Add the `arith_offset` and `ptr_guaranteed_cmp` intrinsics.
- Set `m_get_params` for a generic path inside an expression; without it a const-generic argument there threw `std::bad_function_call` with no diagnostic.

minicargo:
- Don't repoint a build-script *run* dependency at a codegen job: those have no codegen job, so the name is never announced and the whole tree deadlocks.
- Repoint host dependencies at the codegen job under deferred codegen; the old test read as "not already suffixed" but a host job name already ends in "(host)", so its dependencies could be linked against before codegen ran.
- Point a cross-compiled crate's OUT_DIR at the directory its build script actually wrote to - a build script is always a host binary.
- No implicit `foo = ["dep:foo"]` feature when `[features]` names `dep:foo` itself, matching cargo. Otherwise an optional dependency and a same-named ordinary feature merge.
- Accept `crate-type = ["lib"]` as an alias for rlib.
- Report a deadlock with the job each waiter is stuck on, and don't exit 0.

### DIFF
--- a/lib/libproc_macro/src/lex.rs
+++ b/lib/libproc_macro/src/lex.rs
@@ -294,12 +294,6 @@ impl ::std::str::FromStr for TokenStream {
                                 break ;
                             }
                             else if c == '\\' {
-                                // This match had no arms at all, so *every* escape
-                                // panicked - and a panic here kills the proc-macro
-                                // child, which the compiler reports only as
-                                // "Unexpected EOF while reading from child process".
-                                // `indoc`'s `formatdoc!` (used by `instability`, via
-                                // ratatui) emits a string literal containing `\n`.
                                 match some_else!(it.consume() => return err("str eof"))
                                 {
                                 'n' => s.push('\n'),
@@ -331,8 +325,7 @@ impl ::std::str::FromStr for TokenStream {
                                     }
                                     s.push(some_else!(::std::char::from_u32(v) => return err("Invalid `\\u` escape")));
                                     },
-                                // A backslash at end-of-line eats the newline and all
-                                // leading whitespace on the next line.
+                                // A backslash at end-of-line eats the newline and the next line's leading whitespace
                                 '\n' | '\r' => {
                                     while it.next().map(|c| c.is_whitespace()).unwrap_or(false) {
                                         it.consume();
--- a/lib/libproc_macro/src/lex.rs
+++ b/lib/libproc_macro/src/lex.rs
@@ -131,8 +131,10 @@ impl ::std::str::FromStr for TokenStream {
                             '0' => '\0',
                             'n' => '\n',
                             'r' => '\r',
+                            't' => '\t',
                             '\\' => '\\',
                             '\'' => '\'',
+                            '"' => '"',
                             'u' => {
                                 if it.consume() != Some('{') {
                                     return err("Expected `{` after `\\u` in char literal");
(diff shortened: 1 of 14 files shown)


lib/libproc_macro/src/lex.rs
```diff
@@ -292,8 +294,50 @@ impl ::std::str::FromStr for TokenStream {
                                 break ;
                             }
                             else if c == '\\' {
+                                // This match had no arms at all, so *every* escape
+                                // panicked - and a panic here kills the proc-macro
+                                // child, which the compiler reports only as
+                                // "Unexpected EOF while reading from child process".
+                                // `indoc`'s `formatdoc!` (used by `instability`, via
+                                // ratatui) emits a string literal containing `\n`.
```
thepowersgang: This seems more like a review comment, not a good code comment... was this AI?

src/expand/crate_tags.cpp
```diff
@@ -25,6 +25,10 @@ class Decorator_CrateType:
         else if( name == "cdylib" ) {
             crate.m_crate_type = AST::Crate::Type::CDylib;
         }
+        // `Type::ProcMacro` was only reachable from `--crate-type`, not from the inner attribute.
```
thepowersgang: "was" that's not a sensible comment here.

src/hir_typeck/common.cpp
```diff
@@ -756,9 +756,14 @@ ::HIR::LifetimeRef MonomorphiserPP::get_lifetime(const Span& sp, const ::HIR::Ge
...
     {
+    // A lifetime whose index the recorded param list does not cover is passed through rather than aborting, as `monomorph_lifetime` already does for HRTBs.
+    // Lifetimes are erased before codegen, so an unresolved one is inert - which is why `get_type`/`get_value` still assert and these do not.
     case 0:
         if( const auto* p = this->get_impl_params() ) {
-            ASSERT_BUG(sp, lft_ref.idx() < p->m_lifetimes.size(), "Lifetime param " << lft_ref << " out of range for (max " << p->m_lifetimes.size() << ")");
+            if( lft_ref.idx() >= p->m_lifetimes.size() ) {
+                DEBUG("Impl lifetime " << lft_ref << " out of range (max " << p->m_lifetimes.size() << ") - passthrough");
+                return HIR::LifetimeRef(lft_ref.binding);
+            }
```
thepowersgang: ... How valid is this? The logic mostly holds for HRLs, but for named params less so.

thepowersgang (commented):
Main issue is the commenting style.

lib/libproc_macro/src/lex.rs
danifunker: Yes it was AI, I just swept the comments (again) so now everything is describing the current code, not the changes to the code

src/expand/crate_tags.cpp
danifunker: Resolved

src/hir_typeck/common.cpp
danifunker: it now only passes through when the lifetime array is empty, which is the only case I hit. An out-of-bounds index into a populated array still asserts.

PR #418, opened 2026-07-31
Add powerpc-apple-darwin as a target

danifunker:
Builds libcore and libstd for 32-bit PowerPC Mac OS X (10.4/10.5), and produces binaries that run there.

Most of the work is making mrustc's type layout agree with the C compiler's. mrustc emits every Rust type as a plain C struct and lets the C compiler lay it out, then writes `sizeof_assert`/`alignof_assert` typedefs alongside it - so any disagreement is a compile error in the generated C rather than silent corruption, and Darwin/PowerPC's "power" alignment ABI disagrees in four ways:

- A struct member that is not the first has its alignment capped to 4. This has to apply to Rust's own types, not just `repr(C)`. 32-bit PowerPC is the only arch here where it bites, being the only one with 32-bit pointers but 8-byte-aligned u64.
- The cap is on *natural* alignment only. gcc tracks this as TYPE_USER_ALIGN and applies the rule in `place_field` guarded by `if (! DECL_USER_ALIGN (field))`, so a `repr(align(N))` type keeps its alignment however deeply nested, and every aggregate containing it inherits the exemption. `TypeRepr::user_align` models that, propagating outward exactly as gcc does.
- A union takes its *first* member's alignment under this ABI, so codegen_c pins union alignment explicitly rather than letting the C compiler derive it. Members all sit at offset 0, so pinning moves nothing else.
- An align-1 type must be emitted packed even when it is not `repr(packed)`: gcc takes a container's alignment from the member's natural alignment, ignoring the `#pragma pack` the member was defined under, and `aligned(N)` cannot lower alignment to correct it.

A niche enum's size and alignment now come from the final variant layouts rather than from `max_align`, which predates the tag field and so sizes a non-leading payload as though it led. Guarded on the capping ABI; elsewhere the two agree by construction.

Also here:

- `*-apple-darwin` has an empty target_env, as in rustc. Declaring "gnu" made `#[cfg(target_env = "gnu")]` select glibc-specific code on a platform with no glibc. This fixes every apple-darwin target, not just this one.
- 32-bit PowerPC reports 8-byte atomics as available and links libatomic. They are lock-based there, but reporting them unavailable cfg's AtomicU64 out of libcore, which libstd's `sys::unix::time` uses unconditionally on macOS.
- Two libstd patches: `macos_weak` only defines the `fdopendir` weak symbol for x86/x86_64, and the Darwin thread parker needs libdispatch (10.6+), so this target falls through to the generic pthread parker.
- `MINICARGO_NO_DEBUG_ASSERTIONS`, because `debug_assertions` enables `assert_unsafe_precondition!`, which asserts Rust-level invariants a C-ABI-driven layout cannot satisfy - under the alignment cap an 8-aligned field sits at a 4-aligned offset and `ptr::write`'s precondition aborts during `rt::init`. The emitted C is correct; only the assertion disagrees.

This PR needs https://github.com/thepowersgang/mrustc/pull/417 to be merged before it will work.

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -69,7 +69,8 @@ const TargetArch ARCH_POWERPC64LE = {
 const TargetArch ARCH_POWERPC = {
     "powerpc",
     32, true,
-    { /*atomic(u8)=*/true, true, true, false,  true },
+    // 8-byte atomics are lock-based via libatomic here, but still available: cfg'ing out AtomicU64 breaks libstd.
+    { /*atomic(u8)=*/true, true, true, true,  true },
     TargetArch::Alignments(2, 4, 8, 8, 4, 8, 4)
 };
 const TargetArch ARCH_RISCV64 = {
@@ -604,12 +605,13 @@ namespace
                 ARCH_X86_64
                 };
         }
+        // `*-apple-darwin` has an empty target_env (as in rustc); "gnu" selects glibc-only code on a platform with no glibc.
         else if(target_name == "i686-apple-darwin")
         {
             // NOTE: OSX uses Mach-O binaries, which don't fully support the defaults used for GNU targets
             // The first 32bit Intel Mac was Core Solo aka yonah. It allows to use `-march=yonah` like Rust.
             return TargetSpec {
-                "unix", "macos", "gnu", {CodegenMode::Gnu11, false, "x86_64-apple-darwin", {"-march=yonah"}, {}},
+                "unix", "macos", "", {CodegenMode::Gnu11, false, "x86_64-apple-darwin", {"-march=yonah"}, {}},
                 ARCH_X86_64
                 };
         }
@@ -618,7 +620,7 @@ namespace
             // NOTE: OSX uses Mach-O binaries, which don't fully support the defaults used for GNU targets
             // The first 64bit Intel Mac was Core Duo. It allows to use `-march=core2` like Rust.
             return TargetSpec {
-                "unix", "macos", "gnu", {CodegenMode::Gnu11, false, "x86_64-apple-darwin", {"-march=core2"}, {}},
+                "unix", "macos", "", {CodegenMode::Gnu11, false, "x86_64-apple-darwin", {"-march=core2"}, {}},
                 ARCH_X86_64
                 };
         }
@@ -626,15 +628,16 @@ namespace
         {
             // NOTE: OSX uses Mach-O binaries, which don't fully support the defaults used for GNU targets
             return TargetSpec {
-                "unix", "macos", "gnu", {CodegenMode::Gnu11, false, "aarch64-apple-darwin", {}, {}},
+                "unix", "macos", "", {CodegenMode::Gnu11, false, "aarch64-apple-darwin", {}, {}},
                 ARCH_ARM64
                 };
         }
         else if(target_name == "powerpc-apple-darwin")
         {
             // NOTE: OSX uses Mach-O binaries, which don't fully support the defaults used for GNU targets
+            // NOTE: 32-bit PowerPC needs libatomic for the 8-byte atomics (see ARCH_POWERPC)
             return TargetSpec {
-                "unix", "macos", "gnu", {CodegenMode::Gnu11, true, "powerpc-apple-darwin", {}, {}},
+                "unix", "macos", "", {CodegenMode::Gnu11, true, "powerpc-apple-darwin", {}, {}, {"-l", "atomic"}},
                 ARCH_POWERPC
                 };
         }
@@ -642,7 +645,7 @@ namespace
         {
             // NOTE: OSX uses Mach-O binaries, which don't fully support the defaults used for GNU targets
             return TargetSpec {
-                "unix", "macos", "gnu", {CodegenMode::Gnu11, false, "powerpc64-apple-darwin", {}, {}},
+                "unix", "macos", "", {CodegenMode::Gnu11, false, "powerpc64-apple-darwin", {}, {}},
                 ARCH_POWERPC64
                 };
         }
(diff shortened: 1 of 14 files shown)


src/trans/target.cpp
```diff
@@ -2219,6 +2302,14 @@ void Target_ForceTypeRepr(const Span& sp, const ::HIR::TypeRef& ty, TypeRepr rep
...
     set_type_repr(sp, ty, box$(repr));
 }
+bool Target_CapsMemberAlignment()
+{
+    return target_caps_member_alignment();
+}
+bool Target_TypeHasUserAlignment(const Span& sp, const StaticTraitResolve& resolve, const ::HIR::TypeRef& ty)
+{
+    return type_has_user_alignment(sp, resolve, ty);
+}
```
thepowersgang, line 2299: Why do these just wrap an internal function, instead of being the real implementation?

thepowersgang (commented):
Please break into three commits:
- Patch files and script overrides
- The minicargo change
- Codegen changes

Mostly because that makes the git history a little nicer

src/trans/target.cpp
danifunker: No good reason - I factored it out as a local helper, then added a forwarder when codegen_c needed it too rather than just promoting it. Bodies now live in the public functions.

danifunker:
I will be splitting this PR up as per your recommendation now. and will close this PR once everything else is staged.

danifunker:
Commits are split into 3 now, please let me know if there are any other changes I should make to this PR.

Thanks again!

barracuda156:
@thepowersgang Someone should have tagged me here :)

Quick note: avoiding `libdispatch` based on arch is wrong, since it works on 10.6, and 10.6 support ppc (32-bit).

I will compare the ABI stuff with what I use.

barracuda156:
@thepowersgang BTW, my build does not need to disable asserts anymore, provisionally ABI is correct (at least on mrustc level).

barracuda156:
@danifunker Did you actually build `rustc` itself though? There is no publicly available LLVM later than 7.1 that works on powerpc-darwin. You would either need to hack rustc to build with 7.1 or fix LLVM which rustc needs.

barracuda156:
> 10.4/10.5 have no $INODE64 variants, so the plain symbol is the correct one there (from https://github.com/thepowersgang/mrustc/commit/a147c68cfab0dc8dfb58f3db85dc67e3a840bcf7)

@danifunker 10.6 does have `INODE64` variants:

darwin-INODE64

danifunker:
@barracuda156 

I don't have 10.6 running no my G5, so I didn't test that part. My goal is also to target "majority of users" not the 10.6.8. edge case, and my application which was compiled via this process can actually run on a G3 running OSX Tiger (10.4). I didn't build `rustc`, and I don't touch LLVM at all, so the 7.1 ceiling never comes up.

mrustc is a transpiler, not a codegen backend: Rust in, C99 out. What landed in master is the `powerpc-apple-darwin` target spec plus a model of the Darwin/PowerPC alignment rule, so the C mrustc emits has struct layouts that agree with what gcc derives. Everything downstream of that is an ordinary C compile.

Concretely it's a two-machine pipeline:
```
      modern host (fast)              PowerPC Mac, over ssh
      ------------------              ---------------------
      Rust --mrustc--> C99    --->    C99 --gcc10--> PowerPC Mach-O
```
mrustc runs on the modern host (Linux x86_64 here). I point `CC_powerpc_apple_darwin` at a wrapper that *is* the C compiler as far as mrustc is concerned — it rsyncs each translation unit to the Mac, runs gcc there, copies the `.o` back — so minicargo's dependency graph, parallelism and incremental rebuilds keep working across both machines.

The remote compiler is MacPorts **gcc10**, not clang: mrustc's C needs C11 `<stdatomic.h>`, which Xcode's gcc-4.x lacks. Links need three things beyond libSystem — `-latomic` (32-bit PPC has no lock-free 8-byte atomic, so `AtomicU64`'s `__atomic_*_8` calls land there; hence `-l atomic` in the target spec), `-lMacportsLegacySupport` (`pthread_setname_np` is 10.6+), and `-lgcc_s.1` (`_Unwind_GetIPInfo` — in Leopard's `libgcc_s.1` but *not* `libgcc_s.10.4`, so a true 10.4 target needs gcc10's own unwinder or `panic=abort`).

Result: core, alloc, std, panic_unwind, libc, compiler_builtins and test all build for `powerpc-apple-darwin`, and taken a ~380-crate application through to a linked PowerPC Mach-O binary (which is able to run on a minimum of a G3 at 10.4)

**Your point does stand for the bootstrap chain**: `minicargo.mk`' `output/rustc` links `rustc_llvm` and rustc 1.74 wants LLVM 15+, so a  self-hosting `rustc` on powerpc-darwin is blocked exactly as you describe. This is the "run Rust on a PowerPC Mac today, with gcc doing the codegen" route, not the "build rustc there" route. You still need to compile that code on a vintage mac (I used my dual proc G5, similar to yours, but mine is running leopard).

Thanks for the other two notes as well. You're right on `libdispatch` — gating by arch is wrong when 10.6 supports 32-bit ppc; the condition is OS version. And I'd like to see your ABI patch: `MINICARGO_NO_DEBUG_ASSERTIONS` was a workaround for a layout mismatch I never pinned down, and I'd rather delete it than keep it.

barracuda156:
I will make a PR to fix ABI divergence and hopefully other stuff like `libdispatch`, since diverging here will be a pain, we rather get this actually correct instead of approximately correct conditional on specific OS version.

P. S. gcc10 is arguably a bad reference due to the bug in alignments fixed in https://gcc.gnu.org/cgit/gcc/commit/?id=84d080a29a780973bef47171ba708ae2f7b4ee47 (AFAIK not backported to gcc10).
You can have gcc16 from https://github.com/macos-powerpc/powerpc-ports/blob/7b13b07751b7ef2f8673f1a920efb5ff1afab45b/lang/gcc16/Portfile (disclaimer: I do not test on 10.4, so anything 10.4 relies on someone’s else PR and testing).

barracuda156:
@danifunker Besides, thank you for actually pushing this, I would have perhaps postponed with another 6 months of testing, but it is better to have fixes widely available: I guess I am very wrong in assuming ppcports are lol

danifunker:
Sounds awesome. I'm really excited to start seeing actual rust apps on PPC platforms, I think this is a big achievement.
I'm not planning to perform the compilation on 10.4, just use that as a target for the code, so once the C code has been emitted from mrustc. 

I did a check and I'm pretty sure GCC 10.5 (which is what I am running on my g5) has the patch backported in.

barracuda156:
> I'm not planning to perform the compilation on 10.4, just use that as a target for the code, so once the C code has been emitted from mrustc.

On 10.6 everything works natively from pain Xcode to building _some_ Rust apps. Constraints are arch-agnostic mostly: what `mrustc` as such does not handle.
Some examples: https://forums.macrumors.com/threads/i-have-fixed-rust-for-powerpc-via-mrustc-but-it-works-with-real-apps-now.2475548/

> I did a check and I'm pretty sure GCC 10.5 (which is what I am running on my g5) has the patch backported in.

Ah, nice, we only use gcc 10.5 for bootstrap so it is not really updated (until Iain makes his release with collected fixed, including the pending TLS one, which now _is_ crucial).

PR #421, opened 2026-08-08
powerpc-apple-darwin: fix the ABI

barracuda156:
A follow-up to https://github.com/thepowersgang/mrustc/pull/418 that actually matches the ABI correctly (no need to disable asserts anymore) and fixes a few minor bugs.

The `mrustc` build is verified on the real hardware and confirmed to work.
`rustc` bootstrap is still broken, as it was, nothing changes here, but it is broken on unrelated issues (basically on LLVM).

P. S. We still need to address unwinding issue to fix the build on 10.5 (which has no `libunwind` in the system) and fix libs for ppc64 (broken currently).
On 10.6 the working native `mrustc` can be built right now from https://github.com/macos-powerpc/powerpc-ports/blob/c3731ad2d6bb8569e35f8446c134bcadeea02394/lang/mrustc/Portfile and verified via building several Rust ports, see https://github.com/macos-powerpc/powerpc-ports/issues/62 for details.

@danifunker FYI

### DIFF
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -2033,8 +2033,8 @@ namespace {
                             ::std::sort(ents.begin(), ents.end(), sortfn_struct_fields);
                             // - Add tag
                             ents.insert(ents.begin(), Ent());
-                            ents[0].align = tag_size;
-                            ents[0].size = tag_align;
+                            ents[0].align = tag_align;
+                            ents[0].size = tag_size;
                             ents[0].field = ents.size() - 1;
                             ents[0].ty = tag_ty.clone();
 
--- a/src/trans/target.cpp
+++ b/src/trans/target.cpp
@@ -69,8 +69,10 @@ const TargetArch ARCH_POWERPC64LE = {
 const TargetArch ARCH_POWERPC = {
     "powerpc",
     32, true,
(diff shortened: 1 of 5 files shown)


src/trans/target.cpp
```diff
@@ -983,27 +991,11 @@ namespace {
...
-        if( !Target_GetSizeAndAlignOf(sp, resolve, ty, size, align) )
-        {
-            DEBUG("Can't get size/align of " << ty);
-            return false;
-        }
-        out = Ent { idx, size, align, HIR::TypeRef(), false };
-        out.user_align = Target_TypeHasUserAlignment(sp, resolve, ty);
-        out.ty = mv$(ty);
-        return true;
-    }
```
thepowersgang, line 1006: Why was this helper (and its use) removed?

Even if `user_align` is now gone, the rest seemed pretty useful

barracuda156: It was added in powerpc-darwin-specific commit (which does not fully match the ABI), and not used in updated code (or outside of ppc). I do not insist on removing anything, of course; if we make a rewrite that still works correctly, it will be perfectly fine.

barracuda156:
@thepowersgang Thank you for merging!

PR #423, opened 2026-09-05
MIR Lower: don't register drop flags with the `match` pseudo-loop

rautyrauty:
Since 24b3e1d0 (`MIR Lower - Rework scopes for match`) every `match` — and
via HIR every `if` — is lowered inside a Loop scope. `new_drop_flag` attaches
a new flag to the innermost Loop scope so that `complete_scope` re-initialises
it at that scope's entry block.

For a match pseudo-loop that entry block is only reached by paths that enter
*this* match. So a flag created for a variable owned by an enclosing scope —
e.g. a `while let` binding that one arm conditionally moves — is never reset
on the other arms: it keeps the "moved" value from the previous iteration and
the value is never dropped.

Reproducer (`samples/test/match_arm_cond_move_drop.rs`, arm `_ => {}` leaks):

```rust
while let Some(mp) = p.pop() {
    match kinds[mp.idx] {
        0 => { if c { out.push(mp); } }
        1 => { if d { e = Some(mp); } }
        _ => {}
    }
}
```

`-Z dump-mir` shows `drop(_1 IF df$0)` at the end of the loop body while
`df$0 = 1` is only emitted in the entry block of one arm's pseudo-loop.

## Impact

This is what makes the mrustc-built rustc 1.90 exhaust the address space on
32-bit x86 while expanding `rustc_session`'s `options!`: `rustc_expand`'s
`parse_tt_inner` has exactly this shape, so every `MatcherPos` that fails to
match leaks its `Rc<Vec<NamedMatch>>` and every later `Rc::make_mut`
deep-clones. Memory grows cubically in the number of `options!` entries
(40 entries → 142 MB, 80 → 347 MB, 218 → OOM at 3 GB; LLVM-built rustc needs
108 MB). On x86_64 it is only masked by the larger address space.

## Fix

Mark the match pseudo-loop (`new_scope_loop(sp, is_match=true)`) and skip such
scopes when searching for the loop to register a drop flag with. The flag then
lands on the nearest real loop (reset every iteration) or on nothing at all
(function-entry default) — the behaviour before 24b3e1d0.

After the fix `rustc_session` compiles in ~300 MB RSS on i586.

### DIFF
--- a/src/mir/mir_builder.cpp
+++ b/src/mir/mir_builder.cpp
@@ -838,6 +838,9 @@ unsigned int MirBuilder::new_drop_flag(bool default_state)
     {
         if( auto* e = m_scopes.at(m_scope_stack[i]).data.opt_Loop() )
         {
+            // A `match` isn't actually a loop, so promote the drop flag to the parent scope.
+            if( e->is_match )
+                continue;
             e->drop_flags.push_back(rv);
             break;
         }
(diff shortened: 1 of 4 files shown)


src/mir/mir_builder.cpp
```diff
@@ -838,6 +838,15 @@ unsigned int MirBuilder::new_drop_flag(bool default_state)
     {
         if( auto* e = m_scopes.at(m_scope_stack[i]).data.opt_Loop() )
         {
+            // A `match` pseudo-loop is entered once per evaluation, and only
+            // through its own entry block. A flag registered here would be
+            // re-initialised at that block only, so a variable owned by an
+            // enclosing scope that is conditionally moved in one arm keeps a
+            // stale "moved" flag on every path that doesn't enter this match
+            // (e.g. the next iteration of an enclosing loop taking another
+            // arm), and its drop is skipped. Register with the real loop.
```
thepowersgang: Overly verbose comment: It doesn't need the history, just the aim: To promote drop flags to the parent scope (as it's not actually a loop)

rautyrauty: Done.

PR #428, opened 2026-09-08
Solaris 9 & 10 on SPARC, plus three fixes found on the way

danifunker:
This adds two SPARC Solaris targets — sparcv9-sun-solaris (Solaris 10) and sparcv9-sun-solaris2.9 (Solaris 9). The first three commits have nothing to do with Solaris and fix bugs in current mrustc; they're separate commits because they stand on their own.

`minicargo: Allow STD_ENV_ARCH to be overridden`

`std::env::consts::ARCH` was built from the host triple, so a cross-built std reports the wrong architecture — a sparcv9-sun-solaris binary printed arch=x86_64. This affects any cross target, `powerpc-apple-darwin` included. Behaviour is unchanged when the variable isn't set. (The leading component of a triple isn't always the arch name `rustc` uses, so it's overridable rather than derived.)

`expand: Fix `{:02x?}` leaving "?}"` in the formatted output

The specifier emitted a literal `"?}"` into formatted output.

`trans: Fix the emitted signed overflow helpers`

Worth a look regardless of the rest. The `__builtin_mul_overflow_i*` stand-ins mrustc emits for compilers without the type-generic builtins reported overflow for almost any `negative operand: a > INT_MAX/b` says nothing useful when `b` is negative, so `1 * -1` came back as overflowing. Fuzzing the emitted text against GCC's own builtins found 561,004 mismatches across the four signed widths, plus three cases of UB that MSVC tolerates and GCC needn't (INT_MIN/-1 is reachable; add/sub wrapped using signed arithmetic). These helpers are already emitted for MSVC, so this is a live miscompilation today, not something only the new targets reach.

`trans: Add the sparcv9-sun-solaris targets`

Both targets have been exercised by a real application on real hardware, and the two stress different things:

`sparcv9-sun-solaris` built a RustDesk agent for Solaris 10 — X11 capture, VP8 encoding, sockets, threads — which served video to a peer over the network.
`sparcv9-sun-solaris2.9` built a disk-imaging tool for Solaris 9, ~400k lines with its dependencies, whose output is byte-for-byte identical to the same program built by rustc on x86-64 Linux, across three filesystems on both read and write paths.

Both ran on a Sun Blade 2500, which was a Solaris 10 machine when the agent was tested and has since been reinstalled to Solaris 9.

Solaris 9 sits in a gap no package fills. GCC obsoleted the target in 4.9 and deleted it in 5, so 4.9.x is the ceiling — but mrustc's output needs <stdatomic.h> (4.9) and __builtin_*_overflow (5). Rather than require a compiler that cannot exist, emulate-overflow-intrinsics lets a target reuse the suffixed helpers MSVC already had. Two further flags cover a libm predating C99 outright (no INFINITY, no exp2/round/trunc, no float entry points) and a libc without posix_spawn/dirfd.

Three more gaps came with it, each the kind of thing that only surfaces when a target is genuinely new:

#[repr(align(N))] parsed on structs but not unions, erroring with "Unknown union repr 'align'". Solaris' libc declares pad128_t that way, so libc couldn't be parsed at all.
The CC_${TRIPLE} override replaced only -, so a triple naming an OS release produced CC_sparcv9_sun_solaris2.9 — a name no POSIX shell can export, making the documented way to select a C compiler unusable. No triple had contained a . before.
A fieldless #[repr(u32)] enum was passed as a one-field struct across extern "C", where C declares a plain int. SPARC V9 places a 4-byte struct left-aligned in the 8-byte parameter slot and an int in the low half, so the callee read the value shifted 32 bits left and every zstd compression failed with "Unsupported parameter". This is invisible on 32-bit big-endian, where slot and struct are the same width, and on little-endian, where the value is in the low half either way — a 64-bit big-endian target is the first place the two spellings disagree. Arguments only; the return case is documented in the source but untested.

Notes/SolarisSysroot.md covers building a sysroot, which is the one prerequisite that can't be fetched from a public source — Solaris 9 predates OpenSolaris by three years, so no free substitute exists. It documents both routes (a disk image, or install media) and the GCC 4.9.4 build flags a modern host needs.

Claude was used in the development of this feature.


src/trans/codegen_c.cpp
```diff
@@ -259,6 +259,14 @@ namespace {
...
             bool emulated_i128 = false;
             bool disallow_empty_structs = false;
+            // `__builtin_{add,sub,mul}_overflow` arrived in GCC 5. When the target's C
+            // compiler is older than that (or is MSVC, which has never had them) mrustc
+            // emits its own type-suffixed stand-ins instead.
+            bool emulated_overflow_intrinsics = false;
+            // The target's libm predates C99, so mrustc supplies the missing pieces.
+            bool emulated_c99_math = false;
+            // Likewise for a libc missing pieces of POSIX.1-2001/2008.
+            bool emulated_posix2001 = false;
```
thepowersgang, line 269: I think this is probably past the limit of special-casing I'd want to add to mrustc.
Supporting really old GCC versions is going to balloon the backend significantly :(

danifunker:
What would you say the floor is for this project? I really want to support Solaris 10, I can cleanup the PR and target that.

thepowersgang:
Anything that supports C11 (due to the use of `stdatomic.h`) would be a sensible target.

danifunker:
Okay, that opens up Solaris 10 then. Let me update the PR to target that.

PR #431, opened 2026-09-08
minicargo: Add --panic flag passing -C panic=<strategy> to mrustc

rautyrauty:
mrustc falls back to loading panic_abort for any binary that needs a panic_runtime unless the user supplied -C panic=... on the mrustc command line: the `panic_runtime_needed` branch in `src/main.cpp` keys off `params.codegen.panic_type` being empty, and carries a `TODO: Get a panic method from the command line`. minicargo never plumbed that flag through, so every bin target (rustc itself, mrustc's own dump helper, tests) defaulted to panic_abort, regardless of what was sitting in libstd's dep graph. This is what made the previous panic_unwind override useless on rustc_main and what broke `lib/libproc_macro/tools/dump` once the panic_abort dep was deleted from libstd (libpanic_abort.rlib stopped being built, but mrustc still searched for it by name).

Add a CLI option to minicargo and a matching BuildOptions field; pass through to mrustc invocations only (real rustc speaks the flag too but we keep this conservative). minicargo.mk turns it on for the 1.39.0/1.54.0/1.74.0/1.90.0 chain steps where panic_unwind is the intended runtime; the older 1.19.0/1.29.0 paths stay on whatever default they already had.

### DIFF
--- a/tools/minicargo/build.cpp
+++ b/tools/minicargo/build.cpp
@@ -933,6 +933,9 @@ void Job_Build::push_args_common(StringList& args, const helpers::path& outfile,
     if( parent.m_opts.emit_mmir ) {
         args.push_back("-C"); args.push_back("codegen-type=monomir");
     }
+    if( parent.m_opts.panic_strategy && !parent.is_rustc() ) {
+        args.push_back("-C"); args.push_back(format("panic=", parent.m_opts.panic_strategy));
+    }
 
     for(const auto& d : parent.m_opts.lib_search_dirs)
     {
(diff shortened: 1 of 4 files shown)


tools/minicargo/build.cpp
```diff
@@ -933,6 +933,9 @@ void Job_Build::push_args_common(StringList& args, const helpers::path& outfile,
     if( parent.m_opts.emit_mmir ) {
         args.push_back("-C"); args.push_back("codegen-type=monomir");
     }
+    if( parent.m_opts.panic_strategy && !parent.is_rustc() ) {
+        args.push_back("-C"); args.push_back(format("panic=", parent.m_opts.panic_strategy));
+    }
```
thepowersgang, line 938: Why be conservative and not pass to rustc?

If you have a good reason for that, please leave a code comment so future readers can know the reasoning.

rautyrauty: Fixed.

PR #432, opened 2026-09-08
trans/target: Apply repr(simd) alignment rule to struct layout

rautyrauty:
This is necessary because `#[repr(simd)]` was handled in the same `switch` arm as `#[repr(C)]` (no sorting, no packing), so the struct's alignment came from whatever element type happened to be inside. For aarch64 NEON wrappers such as
`core::core_arch::arm_shared::neon::uint8x16_t` (= `Simd<u8, 16>` via the `simd_ty!` alias) that is `[u8; 16]`, giving align=1, while rustc gives align=16. The mismatch made mrustc's const-eval reject zlib-rs's compile-time guard
  const _: () = assert!(core::mem::align_of::<uint8x16_t>() == 16);
and abort with `MIR TODO: panic in constant evaluation` halfway through the aarch64 cargo-build stage of the mrustc bootstrap of rustc-1.90.0.

Compute the alignment from the summed field sizes and feed it through the existing `forced_alignment` channel so both the layout and the emitted C `__attribute__((__aligned__(N)))` agree. This approximates rustc's rule: `llvmlike_vector_align` first consults the target's psABI vector-alignment table and only falls back to `size.next_power_of_two()`, but mrustc has no such table, so the fallback is all we can reproduce.

`repr(simd, packed)` is deliberately left alone. Not because rustc lays it out differently - for a power-of-two lane count `simd_type` takes the same `llvmlike_vector_align` path as the unpacked case - but because mrustc's C backend already emits `#pragma pack(push, 1)` for a packed struct, and forcing a larger alignment on top of that contradicts the pragma and trips every `alignof_assert` for libcore's `core::simd::Simd<T, N>` on i586. Matching rustc there needs the C backend fixed first.

Link: https://github.com/rust-lang/rust/commit/79f41c773aaad6a10c2c752d223bd5aba0371f9b


src/trans/target.cpp
```diff
@@ -1391,10 +1391,24 @@ namespace {
...
+                sorting = StructSorting::None;
+                // Link: https://github.com/rust-lang/rust/commit/79f41c773aaad6a10c2c752d223bd5aba0371f9b
+                // Not for `repr(packed)` - the C backend packs those already.
+                if (str.m_max_field_alignment == 0) {
+                    size_t total_size = 0;
+                    for(const auto& e : ents) total_size += e.size;
+                    size_t simd_align = 1;
+                    while(simd_align < total_size) simd_align <<= 1;
+                    if(forced_alignment < simd_align)
+                        forced_alignment = static_cast<unsigned>(simd_align);
```
thepowersgang: Braces on multi-line if statements please

src/trans/target.cpp
```diff
@@ -1391,10 +1391,24 @@ namespace {
...
             {
             case ::HIR::Struct::Repr::C:
-            case ::HIR::Struct::Repr::Simd:
                 // No sorting, no packing
                 sorting = StructSorting::None;
                 break;
+            case ::HIR::Struct::Repr::Simd: {
+                // No sorting, no packing.
+                sorting = StructSorting::None;
+                // Link: https://github.com/rust-lang/rust/commit/79f41c773aaad6a10c2c752d223bd5aba0371f9b
```
thepowersgang, line 1403: Some context on this link would be nice (i.e. why that particular rustc commit is being refeenced)

thepowersgang (changes requested):

src/trans/target.cpp
rautyrauty: Fixed.

src/trans/target.cpp
rautyrauty: Added two lines: the commit introduced llvmlike_vector_align, and mrustc can only do its next_power_of_two() fallback since there's no psABI vector-alignment table.

PR #434, opened 2026-09-11
trans/codegen_c: Add implicit "memory" clobber to asm! GCC inline-asm

rautyrauty:
Rust's `asm!` is opt-in-restrictive about memory: by default a block may read
or write any memory the surrounding function can see, and only
`options(nomem)` says otherwise ([Reference](https://doc.rust-lang.org/reference/inline-assembly.html#options)).

rustc does the same thing under the same condition ([`asm.rs#L293-L298`](https://github.com/rust-lang/rust/blob/1159e78c4747b02ef996e55082b704c09b970588/compiler/rustc_codegen_llvm/src/asm.rs#L293-L298), introduced in [`abed45ff9fa`](https://github.com/rust-lang/rust/commit/abed45ff9fa3e68f2a32ca12e012f95b9153f4df)).
For LLVM the clobber is decorative — memory effects ride on the call's
`memory(...)` attribute, and the default there is already conservative. For
GCC the clobber *is* the mechanism, so dropping it inverts the meaning.

## Fix

Emit `"memory"` unless `options.nomem` is set.
This costs some optimisation quality around asm blocks.

## What it fixes

`core::num::dec2flt::fpu::fpu_precision::set_precision` guards float parsing
by saving and restoring the x87 control word. It is compiled on i586 —
`#[cfg(all(target_arch = "x86", not(target_feature = "sse2")))]` — and does
not opt out of memory effects:

```rust
asm!(
    "fnstcw word ptr [{}]",
    in(reg) &mut cw,
    options(nostack),
)
```

mrustc emitted it with an empty clobber list, so GCC concluded `cw` was
untouched and kept it in a register at its initialiser. The guard therefore
saves `0`, and its `Drop` writes that back over the real control word.

### End-to-end reproduction

Two i586-native mrustc builds in a sisyphus i586 hasher chroot (gcc 15.3.1),
one at `origin/master` and one with this patch, each used to build libcore
from `rustc-1.90.0-src`. They differ in exactly one place:

```c
/* origin/master */
__asm__ __volatile__(".intel_syntax noprefix; fnstcw word ptr [%0];\n .att_syntax; " : : "r" (var3):);
/* patched */
__asm__ __volatile__(".intel_syntax noprefix; fnstcw word ptr [%0];\n .att_syntax; " : : "r" (var3): "memory");
```

A `no_std` crate calling `f64::from_str("1.2345")`, linked against each
libcore, with a C harness reading the control word around the call:

```
=== patched ===
cw before      = 0x037f
cw after parse = 0x037f
rc=0 v=1.2344999999999999 bits=0x3ff3c083126e978d
post-parse FP op ok, c=0.33333333333333331

=== origin/master ===
cw before      = 0x037f
cw after parse = 0x1475
rc=0 v=1.2344999313354492 bits=0x3ff3c08300000000
post-parse FP op ok, c=0.3333333134651184
```

### Minimal reproduction

The mechanism on its own, no mrustc involved — the emitted line above
compiled as C in the same chroot, once as emitted and once with the clobber:

```c
__attribute__((noinline))
static uint16_t read_cw(void) {
    uint16_t cw = 0;
    uint16_t *var3 = &cw;
    __asm__ __volatile__(".intel_syntax noprefix; fnstcw word ptr [%0];\n .att_syntax; " : : "r" (var3):);
    return cw;                       /* add `: "memory"` above for the other column */
}
```

| | no clobber | `"memory"` |
|---|---|---|
| `-O0` | `0x037f` | `0x037f` |
| `-O1` | `0x0000` | `0x037f` |
| `-O2` | `0x0000` | `0x037f` |
| `-O3` | `0x0000` | `0x037f` |

## Scope

All six `asm!`-derived blocks in libcore (two `fnstcw`, three `fldcw`, one
more) gain the clobber. The three blocks that do not come through
`emit_asm2` — `pause`, `xgetbv`, `cpuid` — are untouched. libcore still
builds.

libcore has no `options(nomem)` block, so the negative case is covered by
inspection of the condition only.

Only x86 is affected in practice: it needs a target where a no-`nomem` block
writes through a pointer and the surrounding code is optimised enough for GCC
to cache across it.


src/trans/codegen_c.cpp
```diff
@@ -5780,9 +5780,16 @@ namespace {
...
                 }
                 m_of << ":";
+                bool need_clobber_comma = false;
                 for(size_t i = 0; i < clobbers.size(); i ++ ) {
-                    if( i > 0 ) m_of << ",";
+                    if( need_clobber_comma ) m_of << ",";
                     m_of << " \"" << clobbers[i] << "\"";
+                    need_clobber_comma = true;
+                }
+                // Link: https://github.com/rust-lang/rust/blob/1159e78c4747b02ef996e55082b704c09b970588/compiler/rustc_codegen_llvm/src/asm.rs#L293-L298
```
thepowersgang: This comment is a bit confusing, probably worth removing it - or just noting the inverted logic between rust's `asm!` and gcc's assembly.

rautyrauty: Done - dropped the link and noted the inverted defaults instead.

thepowersgang (approved):

